PR #3, opened 2013-08-21
Feature: Support for retrieving multiple result sets

WimObiwan:
Necessary when a stored procedure returns multiple result sets.
Useful for e.g. when combining 2 selects in one round trip to the server (SELECT x FROM x; SELECT y FROM y;)


lexicalunit:
This issue was moved to nanodbc/nanodbc#27

PR #4, opened 2013-08-21
Feature: Support for retrieving output/return parameters when calling stored procedure

WimObiwan:
...dures.  You need to retrieve all returned result sets before the output parameters are filled in (using next_result(), provided by separate commit).

Example:
stmt.prepare(conn, "{? = CALL dbo.spMyProc(?,?,?)}");
stmt.bind_parameter(0, &result, NULL, nanodbc::statement::Return);
stmt.bind_parameter(1, str.c_str());
stmt.bind_parameter(2, &i);
stmt.bind_parameter(3, &j, nanodbc::statement::Out);
nanodbc::result resultset = stmt.execute();
while (resultset.next_result()) ;
conn.disconnect();

### DIFF
--- a/nanodbc.cpp
+++ b/nanodbc.cpp
@@ -891,7 +891,7 @@ public:
     }
 
     template<class T>
-    void bind_parameter(long param, const T* value, long* nulls)
+    void bind_parameter(long param, const T* value, long* nulls, param_direction direction)
     {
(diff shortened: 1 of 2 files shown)


lexicalunit:
This issue was moved to nanodbc/nanodbc#28

PR #5, opened 2013-08-21
Fix: Nanodbc exception when execute()-ing a SQL command or stored procedure that doesn't return data

WimObiwan:


lexicalunit:
This issue was moved to nanodbc/nanodbc#29

PR #6, opened 2013-08-26
Added 64-bit support

WimObiwan:
The "long *" pointers (which specify whether or not parameter values are NULL) resulted in compiler errors when using x64 compilation on VC8.  The odbc API expects a SQLLEN which is __int64 on x64 compilations.

To fix this, in nanodbc, I replaced the long with a new null_type that is long on 32-bit, and __int64 on 64-bit.


lexicalunit:
This issue was moved to nanodbc/nanodbc#30

PR #7, opened 2013-10-01
Several changes to improve wide string support.

mcg1969:
First, I noticed that `get_impl<string_type>` was computing the wrong string length for wide strings, so the calls to buffer.resize() were incorrect. I created a NANODBC_STRLEN macro so that `wcslen` is called instead of `strlen` for wide strings.
There was also a problem with `last_error` with wide strings. It's C++'s fault, really, because it's not convenient to have wide string exceptions. So I reworked it to retrieve the wide string from SQLGetDiagRec and crudely convert the result to a narrow string for the exception message. I'm just truncating the 16-bit characters to 8 bits. Conversion to UTF-8 is probably a better choice.
Finally, I changed the parameter number from "long" to "short" to reflect what SQLBindParameter, etc. expect to see.


lexicalunit:
This issue was moved to nanodbc/nanodbc#31

PR #8, opened 2013-10-04
many fixes for windows 64bit unicode

duhlin:
This pull request fixes many issues on win64 unicode with visual studio 2008.
This commit includes the pull request from mcg1969 which solves compilation issue on this plateform.

It has been tested with sqlite3 odbc driver.


lexicalunit:
This issue was moved to nanodbc/nanodbc#32

PR #9, opened 2013-10-14
Change std::snprintf to _snprintf for VC11

spriest487:
Building on VC11 in multi-byte character mode failed with an error about std:: not containing snprintf. Changing this line to match the unicode version a few lines above it fixed the issue.


lexicalunit:
This issue was moved to nanodbc/nanodbc#33

PR #13, opened 2013-11-13
Added bind_null_parameter

DraconPern:
I tested against mysql 5.5.28 on windows.  Worked against timestamp and varchar type.


lexicalunit:
This issue was moved to nanodbc/nanodbc#37

PR #19, opened 2014-02-19
VS2013 and later: set NANODBC_USE_CPP11, silence warnings

mcg1969:
Visual Studio 2013 works great with the NANODBC_USE_CPP11 flag. Why not make that automatic? It still, however, requires the silencing of various warnings in nanodbc.cpp.


lexicalunit:
This issue was moved to nanodbc/nanodbc#44

PR #21, opened 2014-03-05
Example fixes

reubenhwk:
example.cpp should return a proper value in main and it should also print usage info.

Without the usage info, and the check of the command line parameters, users will get a segfault unless they already know how to use example.cpp.


lexicalunit:
This issue was moved to nanodbc/nanodbc#46

PR #25, opened 2014-07-17
Add get_ref to result and adjust unit tests

ctrychta:
I found that when processing large result sets that a lot of time was being spent creating new strings so I added a get_ref function which will fill in the reference with the column value instead of returning it.

While adding unit tests for these changes I ran into a few problems with the existing tests.  I'm not sure how the tests were previously passing so if you think this could somehow be related to something in my environment please let me know.  After these changes I was able to successfully run the sqlite tests on linux and windows.

### DIFF
--- a/nanodbc.h
+++ b/nanodbc.h
@@ -815,7 +815,7 @@ public:
     //!
     //! Columns are numbered from left to right and 0-indexed.
     //! \param column position.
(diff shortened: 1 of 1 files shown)


lexicalunit:
This issue was moved to nanodbc/nanodbc#52

PR #26, opened 2014-07-17
Fixed incorrect columns with multiple result sets

ctrychta:
I found that the current multiple result set handling expected every result set to consist of the same columns so I changed it to rebind the columns for every result set to support different columns in each result set.

I have tested this on linux and windows against MSSQL, but did not see an easy way to write unit tests for this against SQLite.  If you have any ideas please let me know and I will add some tests.

### DIFF
--- a/nanodbc.cpp
+++ b/nanodbc.cpp
@@ -1524,8 +1524,7 @@ public:
 
     ~result_impl() throw()
     {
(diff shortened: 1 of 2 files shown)


lexicalunit:
This issue was moved to nanodbc/nanodbc#53

PR #27, opened 2014-07-21
Added support for SQLProcedureColumns

ctrychta:
This change adds support for the SQLProcedureColumns function.  I have testing on linux and windows against MSSQL, but I don't see an easy way to write unit tests since sqlite doesn't support stored procedures.


lexicalunit:
This issue was moved to nanodbc/nanodbc#54

PR #35, opened 2015-02-13
don't throw exception when failing to set default timeout on stmt

TolonUK:
Some ODBC drivers (e.g. Microsoft Access on Windows) do not support setting a query timeout on a statement (SQL_ATTR_QUERY_TIMEOUT). This change prevents an exception being raised when an error code is returned from attempting to set a timeout of 0 (the default timeout value).


lexicalunit:
LGTM :+1:

lexicalunit:
This issue was moved to nanodbc/nanodbc#62

PR #36, opened 2015-02-26
Wrong string length computation in result::result_impl::get_ref_impl<string_type>()

ChrisBFX:
1. odbc doesn't necessarily add a \0-byte at the end of strings.
col.cbdata_ has the real string length in bytes.
2. wchar_t on gcc is 4-byte while SQLWCHAR is 2-byte


ChrisBFX:
Any comment on this? I think this would also solve https://github.com/lexicalunit/nanodbc/issues/34

My testing platform is Debian8 with unixodbc and mysql 5..5

lexicalunit:
:+1: LGTM, tests pass for me as well. Much thanks!

PR #39, opened 2015-04-23
get<string_type> works for SQL_LONGVARCHAR

manwithahammer:
As per your recommendation in the issue I created, void result::result_impl::get_ref_impl<string_type>(short column, string_type& result) const now handles LONGVARCHAR for SQL_C_CHAR.

### DIFF
--- a/nanodbc.cpp
+++ b/nanodbc.cpp
@@ -2068,13 +2068,29 @@ inline void result::result_impl::get_ref_impl<string_type>(short column, string_
     {
         case SQL_C_CHAR:
         {
-            if(col.blob_)
-                throw std::runtime_error("blob not implemented yet");
-            const char* s = col.pdata_ + rowset_position_ * col.clen_;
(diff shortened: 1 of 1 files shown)


lexicalunit:
Dude, awesome! Have you tested this with your PostgreSQL setup and it works correctly?

manwithahammer:
Yup, tested it against a few different variants of multi-character strings and it seems to work fine. I am also testing against SQL Server, though I don't have any LONGVARCHAR fields to test against right now. FYI, I am using nanodbc to pull SQL query results directly into Excel spreadsheets with a C++ XLL. Works like a charm.

lexicalunit:
Very cool! Glad my library is working out for you. It's been a while since I've worked with C++ and ODBC, but I still remember how bad it was writing straight ODBC code *shudders*. Much thanks for improving the library; I've wanted to tackle CLOB/BLOB support for a long time but just never got around to it.

PR #40, opened 2015-05-12
Enable support for nvarchar(max) and varchar(max) with SQL Server Native Client ODBC driver

kentf:
Hi,

I had a problem with getting any data from a database field where the type was either `varchar(max)` or `nvarchar(max)`, no matter what I tried, it always returned an empty string.

I traced the problem to `auto_bind()`, namely the ODBC call `SQLDescribeCol`.

In my case, the `sqlsize` was 0 for those types. According to the docs, the size could then be between 0 and 2^31-1 bytes.

What worked for me was hooking into the half-done blob functionality, and finish those for the types I use.

As a bonus; I fixed some other compilation issues with C++11 and Unicode and some of the TODOs in the code.

I've tried to adhere to the code style already used, and this works for me. I have only tested this with Win32 and Win64 builds with C++11, non-Unicode and with Unicode. Unfortunately, I haven't been able to test with GCC.

Here is a test case proving the not working broken, and working version working:

```cpp
#include <iostream>
#include <nanodbc.h>

#ifdef NANODBC_USE_UNICODE
# define UNI(a) L ## a
#else
# define UNI(a) a
#endif



int main()
{
	typedef nanodbc::string_type str;

#ifdef NANODBC_USE_UNICODE
	str encoding = UNI("Unicode");
	auto & out = std::wcout;
#else
	str encoding = UNI("Ascii");
	auto & out = std::cout;
#endif


	try
	{
		// Standard output device

		str dsn  = UNI("nanobug");
		str user = UNI("");
		str pass = UNI("");

		nanodbc::connection nanoConnection(dsn, user, pass);

		if (nanoConnection.connected())
		{

			out << UNI("Connected with driver ") << nanoConnection.driver_name() << " with encoding " << encoding << std::endl ;

			std::vector<str> statements
			{
				UNI("DROP DATABASE [nanobug];"),
				UNI("CREATE DATABASE [nanobug];"),
				UNI("USE [nanobug];"),
				UNI("IF OBJECT_ID('simple_test', 'U') IS NOT NULL DROP TABLE simple_test;"),
				UNI("create table simple_test (id int, limited varchar(10), unilimited nvarchar(10), val varchar(max), uni nvarchar(max));"),
				UNI("insert into simple_test values (1, 'one', 'one', 'one', 'one');"),
				UNI("insert into simple_test values (2, 'two', 'two', 'two', 'two');"),
				UNI("insert into simple_test values (3, 'three', 'three', 'three', 'three');")
			};

			for (const auto & stmt : statements)
			{
				out << stmt << std::endl;
				auto result = execute(nanoConnection, stmt);
			}

			str query = UNI("select id, limited, unilimited, val, uni from simple_test");

			nanodbc::statement statement(nanoConnection, query);
			nanodbc::result nanoresults = statement.execute();

			while(nanoresults.next())
			{
				auto id			= nanoresults.get<int>(UNI("id"));
				auto limited	= nanoresults.get<str>(UNI("limited"));
				auto unilimited	= nanoresults.get<str>(UNI("unilimited"));
				auto val		= nanoresults.get<str>(UNI("val"));
				auto uni		= nanoresults.get<str>(UNI("uni"));

				out << 
					UNI("Id: '") << id << 
					UNI("' limited: '") << limited << 
					UNI("' unicode limited: '") << unilimited << 
					UNI("' val: '") << val << 
					UNI("' unicode: '") << uni << UNI("'") 
					<< std::endl;
			}


			int a_mighty_fine_place_to_break=0;
		}

	}
	catch (const std::exception & ex)
	{
		std::cout << ex.what();
	}


	return 0;
}
```

Correct output should be:

```
Connected with driver sqlncli11.dll with encoding Ascii
DROP DATABASE [nanobug];
CREATE DATABASE [nanobug];
USE [nanobug];
IF OBJECT_ID('simple_test', 'U') IS NOT NULL DROP TABLE simple_test;
create table simple_test (id int, limited varchar(10), unilimited nvarchar(10), val varchar(max), uni nvarchar(max));
insert into simple_test values (1, 'one', 'one', 'one', 'one');
insert into simple_test values (2, 'two', 'two', 'two', 'two');
insert into simple_test values (3, 'three', 'three', 'three', 'three');
Id: '1' limited: 'one' unicode limited: 'one' val: 'one' unicode: 'one'
Id: '2' limited: 'two' unicode limited: 'two' val: 'two' unicode: 'two'
Id: '3' limited: 'three' unicode limited: 'three' val: 'three' unicode: 'three'
```

For incorrect output, the `val` and `unicode` values are either empty or full of garbage.

I'm sorry if this will break C++98, I haven't had the capacity to test with a non-C++11 compiler.

### DIFF
--- a/nanodbc.cpp
+++ b/nanodbc.cpp
@@ -16,6 +16,7 @@
 #include <cstring>
 #include <ctime>
 #include <map>
+#include <sstream>
 
 #if defined(NANODBC_USE_CPP11)
     #include <codecvt>
@@ -28,7 +29,7 @@
 #if defined(_MSC_VER) && _MSC_VER <= 1800
     // silence spurious Visual C++ warnings
     #pragma warning(disable:4244) // warning about integer conversion issues.
-    #pragma warning(disable:4312) // warning about 64-bit portability issues.
+    //#pragma warning(disable:4312) // warning about 64-bit portability issues.
     #pragma warning(disable:4996) // warning about snprintf() deprecated.
 #endif
 
@@ -39,6 +40,7 @@
 
 #ifdef _WIN32
     // needs to be included above sql.h for windows
+    #define NOMINMAX
     #include <windows.h>
 #endif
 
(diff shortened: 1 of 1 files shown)


lexicalunit:
Very cool. I'm not terribly interested in maintaining C++98 compatibility moving forward. I just released a 1.0.0 version which will be nanodbc as-is up to before merging in this commit. I'm currently in the process of merging in your commit, stripping out all the CPP11 macros and other junk.

lexicalunit:
Done! Let me know if everything is still working for you and I didn't accidentally introduce a bug when pulling out all the C++03 support cruft. It should be fine, but maybe I made a typo.

kentf:
Hi, thank you very much!

I'll see if this still works tomorrow. If not, I'll fix and create another pull request.

PR #45, opened 2015-09-06
Asynchronous execution

norsync:
Hi,

The third commit adds basic support for asynchronous execution using the notification method (see https://msdn.microsoft.com/en-us/library/hh405038(v=vs.85).aspx), through the addition of two statement member functions, async_execute_direct and async_complete.

I do apologise for the first commit - it was necessary in order to let me compile this in VS2013. I know you're targeting C++11, but you may want to consider this small concession to compatibility, as it's the only thing preventing the use of a development environment that's still pretty widely deployed.

The second commit, setting the ODBC version to 3.8, was necessary in order to make the driver (ODBC Driver 11 for SQL Server) admit that it supports async execution. I imagine you'll want to make this user-modifiable, but I wasn't sure if you'd prefer a compilation option or a runtime option.

Here's a very basic example of usage:

        nanodbc::connection conn(some_connection_string);
        HANDLE hEnv = CreateEvent(NULL, FALSE, FALSE, NULL);
        stmt.async_execute_direct(conn, hEnv, some_statement_string);
        DWORD retcode = WaitForSingleObject(hEnv, INFINITE);
        nanodbc::result row = stmt.async_complete();

If you are interested in supporting async execution but find this implementation unacceptable in one way or another, please let me know how you'd prefer to see it implemented.

Jon Valvatne
Norsync Technology AS

### DIFF
--- a/src/nanodbc.cpp
+++ b/src/nanodbc.cpp
@@ -510,7 +510,7 @@ namespace
                 , rc
                 , env
                 , SQL_ATTR_ODBC_VERSION
-                , (SQLPOINTER)SQL_OV_ODBC3
+                , (SQLPOINTER)SQL_OV_ODBC3_80
                 , SQL_IS_UINTEGER);
             if(!success(rc))
                 NANODBC_THROW_DATABASE_ERROR(env, SQL_HANDLE_ENV);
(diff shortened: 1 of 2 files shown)


lexicalunit:
This is awesome! I like the idea of the user being able to select the ODBC version at runtime or compile time, but it seems like it might be easier to just let it be configured by compile time definition. 3.8 would be a perfectly fine useful default in any case.

I don't mind the changes to support VS2013 as they are minimal and straight forward. I'll do a deeper code review in a few days and get this merged in and release a new version. Thanks again, I'm sure lots of people will find this useful!

PR #51, opened 2015-11-02
Update CMakeLists.txt to enable builds with Visual Studio

mloskot:
Add NANODBC_TEST and NANODBC_INSTALL (default OFF) to allow disabling test and install targets, useful for builds with VS.

Next step, separate PR, is to clean up compilation warnings while building Visual Studio 2015.


lexicalunit:
These changes look good to me. Thanks for improving the state of this project on MSVC!

PR #53, opened 2015-11-04
Define NANODBC_ODBC_VERSION after all the headers are included.

weitjong:
Otherwise, the SQL_OV_ODBC3_80 may not get defined yet and so the previous order would always fallback to ODBC v3.x.

Add another defined check to ensure both 'SQL_ATTR_ASYNC_STMT_EVENT' and 'SQLCompleteAsync' symbols exist before enabling async mode. The header included by the unixODBC could have half-baked support for ODBC 3.8 in some of the Linux distros. In my Fedora 22, the symbol 'SQL_ATTR_ASYNC_STMT_EVENT' exists but the 'SQLCompleteAsync' is not yet defined yet. Without the enhanced conditional check, release v2.7.0 did not build successfully on my Fedora 22.

### DIFF
--- a/src/nanodbc.cpp
+++ b/src/nanodbc.cpp
@@ -23,17 +23,6 @@
     #include <codecvt>
 #endif
 
-// Default to ODBC version defined by NANODBC_ODBC_VERSION if provided.
-#ifndef NANODBC_ODBC_VERSION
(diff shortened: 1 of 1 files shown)


lexicalunit:
Good catch!

PR #55, opened 2015-11-08
Add async connection support, plus extended database_error info.

norsync:
This continues in the vein of pull request #45, this time adding async connection handling through two connection member functions, async_connect and async_complete. It also adds extended information to the database_error object, including the SQLSTATE and the native error code. Here's a basic usage example covering both this and the previous pull request:
```
HANDLE hEnvConn = CreateEvent(NULL, FALSE, FALSE, NULL);
HANDLE hEnvStmt = CreateEvent(NULL, FALSE, FALSE, NULL);

try 
{
    nanodbc::connection conn;
    conn.async_connect(some_connection_string, hEnvConn);
    WaitForSingleObject(hEnvConn, INFINITE);
    conn.async_complete()

    stmt.async_execute_direct(conn, hEnvStmt, some_statement_string);
    WaitForSingleObject(hEnvStmt, INFINITE);
    nanodbc::result row = stmt.async_complete();
}
catch ( nanodbc::database_error e )
{
    std::cout << "Database error: " << e.what() 
              << " (state: " << e.state()
              << ", native error code: " << e.native()
              << ")" << std::endl;
}
```
Please note that this also moves the definition of NANODBC_ODBC_VERSION down below the header inclusions, just like the still-open pull request #53 does. Also, the SQL_API_SQLCOMPLETEASYNC checks added in #53 should probably also be added to the async connection functions added in this pull request.

### DIFF
--- a/src/nanodbc.cpp
+++ b/src/nanodbc.cpp
@@ -23,17 +23,6 @@
     #include <codecvt>
 #endif
 
-// Default to ODBC version defined by NANODBC_ODBC_VERSION if provided.
-#ifndef NANODBC_ODBC_VERSION
-    #ifdef SQL_OV_ODBC3_80
-        // Otherwise, use ODBC v3.8 if it's available...
-        #define NANODBC_ODBC_VERSION SQL_OV_ODBC3_80
-    #else
-        // or fallback to ODBC v3.x.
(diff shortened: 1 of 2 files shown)


lexicalunit:
Very cool! I'm all for more helpful errors. Thank you also for the consistent style and documentation updates. LGTM so I'm merging in these changes now.

PR #57, opened 2015-11-13
Make is_null work after get for drivers that don't correctly indicate null for unbound columns

norsync:
There is currently a bug/limitation in the ODBC drivers for SQL Server, which causes SQLBindCol to never write SQL_NOT_NULL to the length/indicator buffer unless you also bind the data column. This means nanodbc::result::is_null currently returns false for (n)varchar(max) columns that are null.

I don't believe there's a way to properly solve this on nanodbc's end - it is impossible to know if these columns are null or not until you call SQLGetData, and I don't think you want to call SQLGetData inside is_null. But this pull request at least ensures that is_null returns the correct value after SQLGetData is called for that column - ie. after get/get_ref has been called.


lexicalunit:
Mmm, yeah I see what you mean. It is probably worth it to add this information into the documentation for `is_null()`. I'll take care of that. Thanks for the workaround and details!

PR #60, opened 2015-11-23
CMake: fix ODBC linking on Unix

mloskot:
CMake does not guarantee correctly ordered -L and -l flags
if inserted via LINK_FLAGS property.
Fix linking error by passing ODBC libraries via recommended
taret_link_libraries macro.
Bug and fix confirmed on Ubuntu 12.04 with GCC 5.x.


lexicalunit:
Works for me and looks good.

PR #69, opened 2015-11-26
Ensure cbdata_ is always reset in before_move

norsync:
After pull request #57, the result object could potentially be left in a state where it reported is_null true for columns that were not null, which I think is worse than the reverse, so this pull request eliminates that possibility.


norsync:
(By the way, I suspect this leaves the rowset_size_ member of bound_column superfluous.)

lexicalunit:
Yep, agreed on the superfluous `rowset_size_`!

PR #73, opened 2015-11-26
Add build.bat convenience script for Visual Studio users

mloskot:
I suppose, such tiny utility might be useful to those who develop and (re)build nanodbc on Windows frequently. (On Unix, running 'naked' `cmake` usually does the wonders.)

If not that useful, please feel free to close.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,8 @@
 
 /.vagrant
 /doc
-/build*
+/build/
+/build32/
+/build64/
 example
 example.dSYM/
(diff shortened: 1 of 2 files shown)


tjnapster555:
what you think about nuget ?, i use it . its very easy then cmake.

mloskot:
This `build.bat` script is dedicated to those who develop nanodbc on Windows, to simplify (configure-)edit-build-run cycle.

I have been watching NuGet (as well as [CoApp](http://coapp.org/) and their collaboration) with enthusiasm, but such configurations are dedicated more to users of a library.

PR #81, opened 2015-12-08
Add general odbc_test to target variety of ODBC drivers.

### LINKED ISSUE #79: DECIMAL tests rely on broken SQLite ODBC driver
TL;TR: The `decimal_conversion_test` relies on specific behaviour of SQLite ODBC driver and fails for SQL Server, PostgreSQL.

The test uses the following sample data:

```
execute(connection, NANODBC_TEXT("create table decimal_conversion_test (d decimal(9, 3));"));
execute(connection, NANODBC_TEXT("insert into decimal_conversion_test values (12345.987);"));
execute(connection, NANODBC_TEXT("insert into decimal_conversion_test values (5.6);"));
execute(connection, NANODBC_TEXT("insert into decimal_conversion_test values (1);"));
execute(connection, NANODBC_TEXT("insert into decimal_conversion_test values (-1.333);"));    
```

Then, it expects to fetch the **truncated** values for the 2nd and 3rd rows:

```
BOOST_CHECK(results.get<nanodbc::string_type>(0) == NANODBC_TEXT("5.6"));
BOOST_CHECK(results.get<nanodbc::string_type>(0) == NANODBC_TEXT("1"));
```

AFAICT, this relies on the broken behaviour of the SQLite ODBC driver which converts `DECIMAL` column value to `float`.

Inserting the four values above to PostgreSQL and SQL Server preserves decimal places of the actual column type, size, precision:

postgresql-decimal-9-3

sqlserver-decimal-9-3

I can confirm the official PostgreSQL's driver psqlODBC as well as two Microsoft's drivers, *SQL Server Native Client 11* and *ODBC Driver 11 for SQL Server*, correctly preserve the column size/precision.

I can fix that, but I'm not sure which way to choose:
* Switch the test from SQLite to PostgreSQL driver (update Travis CI setup, etc.)
* Detect driver at run-time, somehow, then

   ```
    if (SQLite) {
        <expect broken values
    } else {
        <expect correct values>
    }
   ```

* Move SQLite-specific test from `basic_test.h` to `sqlite_test.cpp`, then add `odbc_test.cpp` targeting non-broken drivers ;)

mloskot:
Add basic utilities to read connection string from `NANODBC_TEST_CONNSTR` environment variable by default.

Add CMake configuration for two separate test targets.

Refactor `sqlite_test` to exercise SQLite-specific behaviours in `decimal_conversion_test`.

Fixes #79.


mloskot:
Uff! The build is finally green :)

@lexicalunit sorry for multiple fixes. Please, let me know what you think. Next, I'll try to run tests on Travis CI.

lexicalunit:
I pulled these changes and tweaked them a bit to fit the test/tests/check paradigm. See the differences here: https://github.com/lexicalunit/nanodbc/compare/mloskot-ml/issue79...lexicalunit/issue79

As well as the commit message, copied here for convenience:

    Tweaks to mloskot-ml/issue79 targets test/tests/check.

    The way the targets were designed is such that:

    - test: runs all tests, but will not build them
    - tests: builds all tests, but does not run them
    - check: builds all tests and then runs all tests

    For individual tests then, it makes sense to use:

    - ${name}_test: runs ${name}_test, but will not build it
    - ${name}_tests: builds ${name}_test, but does not run it
    - ${name}_check: builds ${name}_test and then runs it

mloskot:
Ups, I have missed the actual purpose of the custom `check` target.
Thanks for the corrections.

mloskot:
The [lexicalunit/issue79 build](https://travis-ci.org/lexicalunit/nanodbc/jobs/95700276) is failing due to tests run failure.

I think we can *temporarily* disable the attempt to run tests and just build them, by replacing [.travis.yml:41](https://github.com/lexicalunit/nanodbc/blob/bf0bfd8e73d16f5e8db16dafcba01bf1c05ae8bf/.travis.yml#L41) line
```
make VERBOSE=1 tests
```
with
```
make VERBOSE=1 odbc_test
make VERBOSE=1 sqlite_test
```

mloskot:
I commented your commit with inline comment (see  [4c091d6ae48a8188edf947dd6ede3653c4903e20#commitcomment-14871678](https://github.com/lexicalunit/nanodbc/commit/4c091d6ae48a8188edf947dd6ede3653c4903e20#commitcomment-14871678)) hoping it will get magically auto-linked here, but it isn't.

Copied below:

------


The executable target is named `odbc_tests` (as per the `add_executable` line above).
This line creates custom target named `odbc_test`.
There is also master custom target named `tests`.

Currently, in an IDE, it looks like this:

rename_tests_exe_to_test

What about flipping names of the custom target and the executable, so it is more consistent
 - all custom testing targets would use `tests` (plural).

Then, in an IDE would list the targets (projects) as follows:
```
odbc_test -> executable
odbc_tests -> custom target running the test executable
tests -> custom master target running all tests
```

lexicalunit:
I left some comments on the other thread: https://github.com/lexicalunit/nanodbc/commit/4c091d6ae48a8188edf947dd6ede3653c4903e20#commitcomment-14871678

As for the [.travis.yml:41](https://github.com/lexicalunit/nanodbc/blob/bf0bfd8e73d16f5e8db16dafcba01bf1c05ae8bf/.travis.yml#L41) line, you **definitely do not** want to replace that with

```
make VERBOSE=1 odbc_test
make VERBOSE=1 sqlite_test
```

because that will never build the tests, so it will always fail. It's easier to just do:

```
make VERBOSE=1 check
```

which will build the tests, and then run the tests. No?

lexicalunit:
These changes were resolved and merged via PR #85 :)

Issue #71, opened 2015-11-26
SQLGetStmtAttr with SQL_ATTR_ROW_NUMBER may report SQL_ROW_NUMBER_UNKNOWN


mloskot:
My observations documented in PR #70 may deserve a separate issue report, so here the relevant parts have been copied.

In the `result_impl::position()` function, it may be sensible to check if `SQL_ROW_NUMBER_UNKNOWN` is returned, then use it as the function return value.
Currently, in such case, the function returns arguable value: `SQL_ROW_NUMBER_UNKNOWN - 1 + rowset_position_`.

See my SO question where I asked about implementation differences (I have checked their code) between various drivers: SQLite3, PostgreSQL, MySQL:
[SQLGetStmtAttr output value for SQL_ATTR_ROW_NUMBER](http://stackoverflow.com/q/33940294/151641)

First, the differences and common or possible values need to be confirmed.

Then, also the check in `basic_test::simple_test` might need refinements:
```
BOOST_CHECK(results.position());
```
-----
Reproduced and tested on Windows 10 with Visual Studio 2015 + SQLite ODBC driver (32-bit and 64-bit), building the library for both platforms, 32-bit and 64-bit.

mloskot:
My SO question received answer from someone who seems to be experienced with developing ODBC drviers professionally and the person says:

> The SQLite 3 (SQLiteODBC ) implementation is wrong in this case

PR #82, opened 2015-12-09
Fix result::position to make it consistent with SQL_ATTR_ROW_NUMBER.

mloskot:
If the number of the current row cannot be determined or there is no current row, the driver returns 0.
Otherwise, valid row number is returned, starting at 1.

Handle some incorrect driver implementations (e.g. SQLite ODBC) which instead of 0 return `SQL_ROW_NUMBER_UNKNOWN` (see #71).

Disable test check of `result::position` as not supported by the default `SQL_CURSOR_FORWARD_ONLY`, until cursor type configuration is supported (see #78).

This PR has been tested assuming `odbc_test` is present, so depends on #81.
Also, in order to confirm the fix, I locally enabled `SQL_CURSOR_STATIC` for statement under test.

-----
**NOTE**:
Previous implementation of `result::position` included subtraction of `-1` which had somewhat changed original semantic of `SQL_ATTR_ROW_NUMBER`:
```
return static_cast<unsigned long>(pos) - 1 + rowset_position_;
```
if `pos==0`, it would return overflown "wrapped-around" unsigned value.

This PR removed the `-1` term from the expression.

To summary:
1. I think, `result::position` should conform to semantic of `SQL_ATTR_ROW_NUMBER` exactly and respect the **1-based** numbering of rows. This variant is implemented in this PR.
2. If we still wanted to the custom 0-based numbering, we would have to define a kind `std::string`-like `npos` value to return if the current row can not be determined.

### DIFF
--- a/src/nanodbc.cpp
+++ b/src/nanodbc.cpp
@@ -1938,7 +1938,7 @@ public:
         //
         // NOTE: We try to address incorrect implementation in some drivers (e.g. SQLite ODBC)
         // which instead of 0 return SQL_ROW_NUMBER_UNKNOWN(-2) .
-        if (pos == 0 || pos == SQL_ROW_NUMBER_UNKNOWN)
+        if (pos == 0 || pos == static_cast<SQLULEN>(SQL_ROW_NUMBER_UNKNOWN))
             return 0;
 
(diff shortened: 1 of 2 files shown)


lexicalunit:
LGTM :+1:

PR #85, opened 2015-12-12
RFC: Add convenient access to catalog objects (tables, columns, primary keys)

mloskot:
Here is proposal of catalog feature and I'd like to *request for comments*..

-----

NOTE: this PR has hard dependency on:
1. #81 as it updates the `odbc_test` test.
2. #82 to avoid buggy test failure
3. #83 to avoid buggy test failure too.
So, it mixes in all the commits from those PRs.

All the chnages relevant to this topic are included in the single commit: 7455ccf531eb5e44ce4b883064b59eb9ee45a23c, so please review just this one.

------

Add catalog with basic interface to invoke ODBC API functions `SQLTables`, `SQLColumns`, `SQLPrimaryKeys` via `find_tables`, `find_columns`, `find_primary_keys` respectively.

The catalog provides nested classes: `tables`, `columns`, `primary_keys` corresponding with result sets returned by the ODBC API functions. Those classes work as convenient views through which user can access attributes of the basic database objects.

Add catalog tests (currently, run only against SQL Server).

Note, the `catalog` class is a friend of the `result` class, becasue the former requires access to the private constructor.

-----

Travis CI status is red, but it is due to attempt to run tests while ODBC and target database are not installed completely.


lexicalunit:
Actually, I just pulled this and did some testing and it looks like this incorporates the changes I made in [lexicalunit/issue79](https://github.com/lexicalunit/nanodbc/compare/lexicalunit/issue79), so I'm good with this PR as it is :+1: 

I'm not sure why the build failed though... I'm gonna merge this now and keep it on master (not make a release yet) and fix things in a subsequent commit.

PR #88, opened 2015-12-18
Adds support for while not end iteration.

lexicalunit:
Partly addresses issue #78 to provide `while(!r.end())` style iteration.


mloskot:
Yes, looks good.
I like the idea of managing the *at the end* state internally, in cursor type agnostic manner.
I prefer forward-only cursors in most cases.

Thanks!

PR #90, opened 2016-01-15
Migrates tests from Boost.Test to Catch framework.

lexicalunit:
Move away from Boost.Test; use the much nicer Catch instead.

PR #93, opened 2016-01-29
Restore ${ODBC_LIBRARIES} in target_link_libraries arguments.

mloskot:
It was removed in #90 (accidentally?), but it is required to enable successful linking against ODBC libraries on Windows.


lexicalunit:
:+1:  Thanks!

PR #94, opened 2016-01-29
Propose more portable test for a table presence.

mloskot:
Remove non-standard `IF EXISTS` syntax from `DROP TABLE`.
Use query based check if a table exists.

Since the `odbc_test.cpp` is a generic test, it may be a good idea to keep it clear from non-standard SQL statements.


lexicalunit:
Good idea :+1:

PR #96, opened 2016-01-30
Refactors code to use u16string instead of wstring.

lexicalunit:
Resolves issue #86.

### DIFF
--- a/src/nanodbc.cpp
+++ b/src/nanodbc.cpp
@@ -21,8 +21,8 @@
 
 // User may redefine NANODBC_ASSERT macro in nanodbc.h
 #ifndef NANODBC_ASSERT
-#include <cassert>
-#define NANODBC_ASSERT(expr) assert(expr)
+    #include <cassert>
+    #define NANODBC_ASSERT(expr) assert(expr)
 #endif
 
 #ifdef NANODBC_USE_BOOST_CONVERT
@@ -80,7 +80,7 @@
         #define NANODBC_STRFTIME std::wcsftime
         #define NANODBC_STRLEN std::wcslen
         #define NANADBC_STRNCMP std::wcsncmp
-        #define NANODBC_UNICODE(f) f ## W
+        #define NANODBC_FUNC(f) f ## W
         #define NANODBC_SQLCHAR SQLWCHAR
     #else
         #define NANODBC_TEXT(s) s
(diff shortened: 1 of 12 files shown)


lexicalunit:
This PR switches everything over to `u16string` and strips out `wstring` and `wchar_t` everywhere. It should make unicode work on *nix builds now and leave Windows builds unaffected. There's a lot of changes here though. For example the `swprintf()`, `wcsftime()`, `wcslen()`, `wcsncmp()` family of functions are completely gone now. Instead, we narrow or widen strings as needed and do any manipulation with `string` rather than `u16string`. String streams are removed as they're not implemented for `u16string`.

Would love to get your thoughts and tests on a Windows system @mloskot.

mloskot:
On Windows, the new implementation hits a bug in VS2015 (including Update 1):

```
nanodbc.lib(nanodbc.obj) : error LNK2001: unresolved external symbol
"__declspec(dllimport) public: static class std::locale::id std::codecvt<char16_t,char,struct _Mbstatet>::id"
(__imp_?id@?$codecvt@_SDU_Mbstatet@@@std@@2V0locale@2@A)
```

The error is caused by the line like this (`nanodbc.cpp` and `example_unicode_utils.h`):
```
std::wstring_convert<std::codecvt_utf8_utf16<char16_t>, char16_t>().from_bytes(in);
```

In discussion here, [VS 2015 RC linker std::codecvt error](https://social.msdn.microsoft.com/Forums/en-US/8f40dcd8-c67f-4eba-9134-a19b9178e481/vs-2015-rc-linker-stdcodecvt-error), Stephan T. Lavavej confirms the bug is still in the VS2015 Update 1, and recommends:

> Use wchar_t or unsigned short instead of char16_t, and it should work.

mloskot:
I have tested the workaround #107 with SQL Server and it seems working fine. I hope I haven't missed any additional hacks :)

lexicalunit:
I broke this PR up into 9 individual sequential commits, each one pertaining to one logical change in the code. It should make the changes easier to reason about :)

mloskot:
:+1:

lexicalunit:
Yay, merging!

src/nanodbc.cpp
```diff
@@ -2596,81 +2595,107 @@ inline void result::result_impl::get_ref_impl<string_type>(short column, string_
...
-                , column_size + 2
-                , NANODBC_TEXT("%.*lf")         // restrict the number of digits
+            std::string buffer;
+            const SQLULEN width = column_size + 2; // account for decimal mark and sign
+            buffer.reserve(width + 1); // ensure terminating null
+            buffer.resize(buffer.capacity());
+            using std::fill;
+            fill(buffer.begin(), buffer.end(), '\0');
+            const double data = *reinterpret_cast<double*>(col.pdata_ + rowset_position_ * col.clen_);
+            printf("data: %.*lf\n", col.scale_, data);
```
mloskot: It looks like printf-based debugging left over, doesn't it?

lexicalunit: Yep, will remove.

PR #97, opened 2016-01-30
Avoid swallowing database_error if DROP TABLE failed.

mloskot:
Refines previous change (#94), so failed attempt to drop existing table is signalled to user with meaningful error.


lexicalunit:
Ah yeah, this makes sense.

PR #98, opened 2016-01-30
Create view based on test table for catalog lookup test.

mloskot:
Use custom created view for basic view lookup test - this makes the test more portable across various databases.

Run the test exercising view lookup within a schema against **SQL Server** only as it relies on known schema, namely `INFORMATION_SCHEMA`.

Fiddling with portable schema creation might lead to overly complicated tests


lexicalunit:
:+1:

PR #99, opened 2016-01-30
Fix catalog::columns::is_nullable to handle valid NULL.

mloskot:
`SQLColumns` is allowed to report [IS_NULLABLE](https://msdn.microsoft.com/en-us/library/ms711683.aspx) as `NULL`.


lexicalunit:
Nice catch!

PR #100, opened 2016-01-30
Fix SELECT result sorting with NULL values involved.

mloskot:
Since the SQL standard does not define default sort order of `NULL` values, it is DBMS-specific, so the test is giving different results with **SQL Server** vs **PostgreSQL** (and presumably other DBMS).

Hence, the test needs to use predictable and reproducible ordering on every ru


lexicalunit:
Good catch here, LGTM.

PR #103, opened 2016-01-31
SQL_FLOAT size depends on numeric precision radix.

mloskot:
Some drivers report digits, some report bits, so the column size needs to be interpreted accordingly.


lexicalunit:
I had a feeling this test wasn't perfect. Thanks for this fix :+1:

PR #104, opened 2016-01-31
Check SQL_WVARCHAR along SQL_VARCHAR data type.

mloskot:
Depending on Unicode support, drivers may report the same SQL character column using one of the two variations.

Tested with SQL Server and PostgreSQL.


lexicalunit:
Thanks for testing this and the fix. LGTM.

PR #106, opened 2016-01-31
Fix retrival of variable-length data in parts.

mloskot:
The termination condition of the loop in which `SQLGetData` is called to retrieve varlen data in chunks was incomplete and incorrect. It only tested `SQLGetData` return code is **greater than 0**, what led to infinite loop in case the call returned `SQL_NO_DATA(100)`.

The issue has been reproduced number of times while querying variable-length data from SQL Server,
and the MSDN confirms `SQL_NO_DATA` is a valid return code for `SQLGetData`. So, it should be handled properly.

Test for this fix is not included, but it needs to be added soon.


lexicalunit:
:+1:

PR #107, opened 2016-01-31
Work around VS2015 bug in std::codecvt for char16_t

mloskot:
Attempt to use of `std::wstring_convert` with `char16_t` causes linker errors due to [bug confirmed in VS2015](https://social.msdn.microsoft.com/Forums/en-US/8f40dcd8-c67f-4eba-9134-a19b9178e481/vs-2015-rc-linker-stdcodecvt-error) (including Update 1).

This patch applies workaround specific to VS2015, based on intermediate casting of string elements to/from `int16_t`.

Fixes PR #96.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,14 +18,14 @@ language: cpp
 compiler:
   - gcc
 env:
-  - DB=mysql USE_UNICODE=OFF USE_BOOST_CONVERT=OFF
-  - DB=mysql USE_UNICODE=OFF USE_BOOST_CONVERT=ON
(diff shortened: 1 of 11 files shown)


lexicalunit:
:+1: If it's the recommended solution given by Stephan T. Lavavej then I'm inclined to agree with it. Let's merge this PR once #96 is merged.

mloskot:
Stephan only suggest to use *unsigned short instead of char16_t* and does not present exactly how the workaround should look like. So, this patch is based on derivative and/or an obvious workaround.

lexicalunit:
Merged into `unicode_overhaul` :+1:

PR #110, opened 2016-01-31
Add example program table_schema.cpp.

mloskot:
The new example shows how to query table schema details.

Add any_to_string wrapper on top of convert util to handle any type in generic manner.

Tested against SQL Server and PostgreSQL.


lexicalunit:
LGTM

PR #118, opened 2016-03-21
Fixes implementation of binary blob.

lexicalunit:


lexicalunit:
Issue #117 prompted me to look into how nanodbc handles fetching of blob types. I found a number of issues:

- We need to make sure to start with an empty buffer before each call to `SQLGetData()`.
- `SQL_BINARY` should be considered a BLOB datatype.
- When calling `SQLGetData()`, we should pass `col.ctype_` for the target type.
- The buffer length argument should be one less than the allocated buffer size to prevent buffer overruns.

I have not been able to test these changes on MariaDB yet, but they are now working on SQLite and MySQL. New functional tests have been added to prevent a regression.

mloskot:
@lexicalunit I missed this PR.

FYI, I have been playing with binary support as well, in [mloskot/example-select-binary](https://github.com/mloskot/nanodbc/tree/ml/example-select-binary) branch. There are two commits:
* the feature 486a2b729b0d54646b418c65c643b66ee00cb525
* some testing codes 028c6b76108404b32bfa0b16b6b98d7ed21fdd32

A will compare my impl. with yours and see if there is any need to update.
Perhaps you could have a look at mine too.

lexicalunit:
:+1: I'll try to take a look at it soon. I'm really swamped at work for the next few months. Crunch time.

mloskot:
@lexicalunit No worries, that is nothing special really, just reading into array of bytes, instead of string type. Instead, I'm trying to improve the current `SQL_C_BINARY` support.

PR #129, opened 2016-04-13
Fix copying of buffer to output string for SQL_C_BINARY

mloskot:
Binary data might contain null bytes, thus it is necessary to pass length of buffered data into `std::string::append`.

Otherwise, output data might be truncated (issue confirmed with SQL Server BINARY values).
Add similar fix for `SQL_C_WCHAR` blob for similar reasons.

Additionally, if `SQLGetData` returns none of expected codes (e.g. `SQL_ERROR`), throw `database_error` with diagnostic info.

### DIFF
--- a/src/nanodbc.cpp
+++ b/src/nanodbc.cpp
@@ -2519,6 +2519,8 @@ inline void result::result_impl::get_ref_impl<string_type>(short column, string_
             {
                 // Input is always std::string, while output may be std::string or wide_string_type
                 std::string out;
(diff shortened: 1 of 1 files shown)


mloskot:
Ups, this is wrong. Please, bare with me, I'm correcting this PR.

mloskot:
The PR is ready for review.

lexicalunit:
Oh, good catch here! LGTM 👍

PR #130, opened 2016-04-13
Add support to get binary data as array of bytes

mloskot:
Currently, `SQL_C_BINARY` can be retrieved into `nanodbc::string_type`, what is not optimal especially in Unicode mode.
Propose `vector<uint8_t>` as universal binary data container.
Add overload of `result_impl::get_ref_impl` for `vector<uint8_t>`.

Add test case for SQL Server binary data retrieval.
Add to_hex_string utility to the base fixture.

-----

For `NANODBC_USE_UNICODE` builds, it depends on PR #129


lexicalunit:
I like this idea but the build failed for postgresql. Looks like it doesn't have a `BLOB` datatype. I think we should be using `BYTEA` instead.

mloskot:
Generally, the `odbc_test.cpp` is more an example than practically usable test - unless we figure out SQL abstraction to handle SQL syntax differences.
So, I'm going to derive `postgresql_test.cpp` as we do for MySQL and others.

lexicalunit:
Looks like tests are passing now 👍

PR #134, opened 2016-04-15
Prefer typedef std::wstring string_type on Visual C++ with Unicode

### LINKED ISSUE #133: Q: Keep typedef std::wstring string_type on Visual C++ with Unicode?
This is related to PR #96.

Would it be possible to stick to `typedef std::wstring string_type` while building with Visual C++ on Windows with `NANODBC_USE_UNICODE=ON`?

Shortly, it would mean the following setup:
```
#ifdef NANODBC_USE_UNICODE
    #ifdef _MSC_VER
        typedef std::wstring string_type;
        typedef std::wstring wide_string_type;
        #define NANODBC_CODECVT_TYPE std::codecvt_utf8_utf16
        #define NANODBC_TEXT(s) L ## 
    #endif
#endif
```

On Windows, for Visual Studio/C++ Unicode build (default), `wchar_t`, `std::wstring`, `L` are de-facto standard yet convenient Unicode string facilities. Windows makes a homogeneous platform in this regard, so I believe it is safe to assume that simply:

 _Visual Studio + Unicode_ => `wchar_t`.

If this proposal receives agreement, I will provide a patch and test it with the main four DBMS on both, Windows and Linux.

mloskot:
Closes #133 

------

I have tested with VS2015 + SQL Server running the `mssql_test.cpp` proposed in #139.
I need to rely on Travis CI to check others for me.

### DIFF
--- a/examples/example_unicode_utils.h
+++ b/examples/example_unicode_utils.h
@@ -7,18 +7,28 @@
 #include <locale>
 #include <string>
 
+#ifdef NANODBC_USE_IODBC_WIDE_STRINGS
+#error Examples do not support the iODBC wide strings
+#endif
+
 #ifdef NANODBC_USE_UNICODE
     #undef NANODBC_TEXT
-    #define NANODBC_TEXT(s) u ## s
+    #ifdef _MSC_VER
+        #define NANODBC_TEXT(s) L ## s
+    #else
+        #define NANODBC_TEXT(s) u ## s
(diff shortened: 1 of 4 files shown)


lexicalunit:
So long as tests are passing on Travis CI I think we're safe to merge this for the connivence of Windows users. I read through the code and it seems reasonable to me. 👍

PR #140, opened 2016-05-05
Resolves unexpected bind() with nulls set to nullptr behavior.

### LINKED ISSUE #122: Unexpected statement::bind_param behavior when nulls parameter is nullptr
Hi

statement class has several overloads of bind method:
1. `void bind(short, const T*, std::size_t , param_direction);`
2. `void bind(short, const T*, std::size_t , const bool* nulls, param_direction direction);`

I've faced with unexpected behavior of  second function. If nulls parameter is equal to nullptr then implementation binds parameter as NULL value. This happens because of following code:
statement::statement_impl::bind()
        `if((null_sentry && !equals(values[i], *null_sentry)) || (nulls && !nulls[i]))`
           `bind_len_or_null_[param][i] = parameter_size;`

Maybe it's by design but I would expect that providing nullptr means 'I don't care about null' and this overload is just more generalized version of first function.
i.e. bind(0, &x, 1, nullptr, PARAM_IN) is equal to bind(0, &x, 1, PARAM_IN);

lexicalunit:
Closes #122.

### DIFF
--- a/src/nanodbc.cpp
+++ b/src/nanodbc.cpp
@@ -1826,7 +1826,7 @@ void statement::statement_impl::bind(
     prepare_bind(param, elements, direction, data_type, param_type, parameter_size, scale);
 
     for(std::size_t i = 0; i < elements; ++i)
-        if((null_sentry && !equals(values[i], *null_sentry)) || (nulls && !nulls[i]))
+        if((null_sentry && !equals(values[i], *null_sentry)) || (nulls && !nulls[i]) || !nulls)
             bind_len_or_null_[param][i] = parameter_size;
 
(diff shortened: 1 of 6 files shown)


lexicalunit:
I think it'd be good to prevent Travis CI from kicking off two builds for a PR. Right now it seems to create one for the PR and one for the topic branch. Ideally it'd only kick one off for the PR, and just ignore the topic branch.

lexicalunit:
Rebased off of master w/ the branch whitelist configuration added for CI. Should hopefully make the CI cycles faster and reduce load on Travis and AppVeyor.

PR #141, opened 2016-05-05
Prevent Travs CI from building topic branches without PRs.

lexicalunit:
I believe this should prevent Travis CI from building anything except for the specific branches listed and pull requests. We won't get builds for topic branches anymore, which I think is fine. To initiate a build we'll have to open a PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,8 @@
+branches:
+  only:
+    - master
+    - latest
+    - release
 addons:
   apt:
     sources:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -6,6 +6,12 @@ platform: x64
 
 configuration: Release
 
+branches:
+  only:
+    - master
+    - latest
+    - release
+
 matrix:
   fast_finish: false # set this flag to immediately finish build once one of the jobs fails.
 


lexicalunit:
Seems like the new configuration is working here. We only got the one Travis CI build for this PR, instead of one for the PR and this topic branch :)

lexicalunit:
Sweet. A similar configuration for AppVeyor was straightforward.

lexicalunit:
Gonna merge this now because it will help make CI more responsive asap.

mloskot:
:+1: 

FYI, I think the build matrix could be simplified by removing the `DB=<name>` variable form both, Travis CI and AppVeyor. Then every build job would run all tests against all DBMS - required services run on every build container anyway,

lexicalunit:
Tho, it is kinda nice to have each DB broken out into it's own build. It helps when reading console output cause you know that everything in a given console output is related to that on DB. There's no confusion about which DB you're looking at. But... I don't think it'd be that confusing to have all the tests for all the DB run in one build and it'd definitely save time to do it that way. I'd be willing to experiment with it and see how we feel about it @mloskot 👍

mloskot:
Yes, DB broken out matrix looks good.
Let's wait and see if we ever get annoyed by too long build waiting times :)

PR #144, opened 2016-05-06
Fix overflowing transaction counter

### LINKED ISSUE #143: Transactions counter overflows
Is this intentional the `connection_impl` tracks number of transactions 'backwards' ?

Since commit 999faad058c5be31d6c8d15f015349d801b503ca, the reference counter is implemented this way:

```
    std::size_t ref_transaction()
    {
        return --transactions_;
    }

    std::size_t unref_transaction()
    {
        return ++transactions_;
    }
```

and used as follows (notice the comments 1. and 2.)"

```
transaction_impl(const class connection& conn)
    : conn_(conn)
    , committed_(false)
    {
        ...
                                  // 1.   transactions_  == 0 
        conn_.ref_transaction();  // 2. --transactions_  => size_t(-1)
    }
```

Before the changes in 999faad058c5be31d6c8d15f015349d801b503ca, `transaction` class constructor (equivalent to the current `transaction_impl` constructor above) counted transactions in the 'natural' increasing order:

```
++conn_.transactions_;
```

It smells to me that the implementation of current `ref_transaction` and `unref_transaction` methods has been mixed up.

mloskot:
Correct mixed up definitions of ref_/unref_transaction.
Update transaction test to track the counter value.

Fixes #143

### DIFF
--- a/src/nanodbc.cpp
+++ b/src/nanodbc.cpp
@@ -994,15 +994,16 @@ public:
         return string_type(&name[0], &name[strarrlen(name)]);
     }
 
-
     std::size_t ref_transaction()
     {
-        return --transactions_;
+        return ++transactions_;
     }
 
     std::size_t unref_transaction()
     {
-        return ++transactions_;
+        if(transactions_ > 0)
+            --transactions_;
+        return transactions_;
     }
 
     bool rollback() const
--- a/test/base_test_fixture.h
+++ b/test/base_test_fixture.h
@@ -925,30 +925,44 @@ struct base_test_fixture
 
         check_rows_equal(execute(connection, query), 10);
 
+        REQUIRE(connection.transactions() == 0);
         {
             nanodbc::transaction transaction(connection);
+            REQUIRE(connection.transactions() == 1);
             execute(connection, NANODBC_TEXT("delete from transaction_test;"));
             check_rows_equal(execute(connection, query), 0);
+            REQUIRE(connection.transactions() == 1);
             // ~transaction() calls rollback()
         }
+        REQUIRE(connection.transactions() == 0);
 
         check_rows_equal(execute(connection, query), 10);
 
+        REQUIRE(connection.transactions() == 0);
         {
             nanodbc::transaction transaction(connection);
+            REQUIRE(connection.transactions() == 1);
             execute(connection, NANODBC_TEXT("delete from transaction_test;"));
             check_rows_equal(execute(connection, query), 0);
-            transaction.rollback();
+            REQUIRE(connection.transactions() == 1);
+            transaction.rollback(); // only requests rollback performed in ~transaction()
+            REQUIRE(connection.transactions() == 1); // transaction not released yet
         }
+        REQUIRE(connection.transactions() == 0);
 
         check_rows_equal(execute(connection, query), 10);
 
+        REQUIRE(connection.transactions() == 0);
         {
             nanodbc::transaction transaction(connection);
+            REQUIRE(connection.transactions() == 1);
             execute(connection, NANODBC_TEXT("delete from transaction_test;"));
             check_rows_equal(execute(connection, query), 0);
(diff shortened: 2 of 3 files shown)


test/base_test_fixture.h
```diff
@@ -871,30 +871,44 @@ struct base_test_fixture
...
+        REQUIRE(connection.transactions() == 0);
         {
             nanodbc::transaction transaction(connection);
+            REQUIRE(connection.transactions() == 1);
             execute(connection, NANODBC_TEXT("delete from transaction_test;"));
             check_rows_equal(execute(connection, query), 0);
-            transaction.rollback();
+            REQUIRE(connection.transactions() == 1);
+            transaction.rollback(); // only requests rollback performed in ~transaction()
+            REQUIRE(connection.transactions() == 1); // transaction not released yet
```
mloskot: This is another unclear aspect of the `transaction_impl`: Is this intentional that `rollback()` call does not perform the actual `SQLEndTran(SQL_ROLLBACK)` ?

Currently, there is lack of symmetry in transaction state (counter too, relates to #143) between the `commit` and the `rollback` actions.

lexicalunit: The intent is that `rollback()` marks the transaction for rollback, but that the actual rollback only actually occurs when the transaction object is cleaned up (via it's destructor; i.e. RAII).

src/nanodbc.cpp
```diff
@@ -994,15 +994,15 @@ class connection::connection_impl
...
     std::size_t ref_transaction()
     {
-        return --transactions_;
+        return ++transactions_;
     }
 
     std::size_t unref_transaction()
     {
-        return ++transactions_;
+        NANODBC_ASSERT(transactions_ != 0);
```
lexicalunit: Is there a benefit to not asserting here? Perhaps if `transactions_` already equals `0` we can just return?

lexicalunit:
Added two comments but otherwise this PR LGTM!

test/base_test_fixture.h
mloskot: Althouth the RAII makes sense, my intuition expected the symmestry with the `commit`. Actually, the `commit` doc is incorrect:

```
    //! \brief Marks this transaction for commit.
    void commit();
```
It should rather read "Commits transaction immediately".

I don't request any changes. If, however, we wanted to improve the current interface, I would suggest to use named rollback modes:

```
enum class RollbackPoint { ScopeExit, Immediate };
void rollback(RollbackPoint when = ScopeExit);
```

Boolean parameters, especially with default values, make code less readable.
Specifically, if I saw `rollback(false)` my :eyes: could have rolled :)

src/nanodbc.cpp
mloskot: If I changed the code to `if (transactions_ > 0) --transactions_;` it would be
* change in the original behaviour.
* unrelated to this particular bug fix

both I prefered to avoid and I used assertion to *comment* some obvious expectations.

I will replace it with the `if-condition`.

mloskot:
Replaced the assertion in `unref_transaction` with the condition:
```
if(transactions_ > 0)
    --transactions_;
return transactions_;
```

test/base_test_fixture.h
lexicalunit: I've been writing too much Python lately. I wanted to write something like `rollback(delayed=False)` then I realized I couldn't do that in C++ :(

mloskot: > A transaction that is destroyed without first having commit() called on it will default to
> rolling back, so there's no need to manually call rollback()
> that I can think of off the top of my head.

@lexicalunit I totally agree the `rollback` is redundant. Since it might be confusing (my notes on symmetry above), it could be removed or marked as deprecated. Shall we have a separate PR?

lexicalunit: Yep, separate PR sounds good to me 👍

PR #145, opened 2016-05-06
Add test for PostgreSQL

mloskot:
Set up PostgreSQL testing on Travis CI and AppVeyor.
Comment on `SQL_FLOAT` column size (15 or 17 digits) observed in different environments.

------

As pointed in https://github.com/lexicalunit/nanodbc/pull/130#issuecomment-217127673, generic `odbc_test.cpp` serves more as an example and we still need to maintain sets of DBMS-specific test cases.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,10 +15,12 @@ addons:
       - libmyodbc
       - libsqliteodbc
       - mysql-client
+      - odbc-postgresql
       - unixodbc
       - unixodbc-dev
 services:
   - mysql
+  - postgresql
 language: cpp
 compiler:
   - gcc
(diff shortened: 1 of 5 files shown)


lexicalunit:
LGTM! Too bad that postgresql test doesn't work on Linux. I don't think there is actually a viable postgresql-odbc-connector we can use with unixODBC for this so we may just want to disable to the tests  on travis ci for now.

mloskot:
I have no idea what is going wrong with the DSN-less connection using unixODBC + psqlODBC on Travis CI. Simply, it should work1

I've been trying to reproduce the problem locally, but having some Vagrant issues with my Linux-based box. So, it is being investigated but will take a while.

In the meantime, I will disable the test on Travis CI.

mloskot:
Haha, I think I found why unixODBC + psqlODBC connections were failing on Travis CI.
Can anyone spot the difference, between this connection string:

```
$ example_table_schema "DRIVER={PostgreSQL ANSI};Server=127.0.0.1;Database=nanodbc;UID=vagrant;PWD=vagrant;" myt
table_catalog            : nanodbc
table_schema             : public
table_name               : myt
--------------------------------
ordinal_position         : 1
column_name              : xid
type_name                : int4
data_type                : SQL_INTEGER
sql_data_type            : SQL_INTEGER
sql_datetime_subtype     : 0
column_size              : 10
decimal_digits           : 0
buffer_length            : 4
char_octed_length        : -1
numeric_precision_radix  : 10
nullable                 : 1
is_nullable              :
remarks                  :
column_default           :
```
and this one?

```
$ example_table_schema "Driver={PostgreSQL ANSI};Server=127.0.0.1;Database=nanodbc;UID=vagrant;PWD=vagrant;" myt
/home/vagrant/nanodbc/src/nanodbc.cpp:863: IM002: [unixODBC][Driver Manager]Data source name not found, and no default driver specified
```

:)

mloskot:
@lexicalunit FYI, Travis CI now runs PostgreSQL tests, since https://travis-ci.org/lexicalunit/nanodbc/jobs/129087139

PR #148, opened 2016-05-10
Handle SQL_SS_UDT data as binary

mloskot:
Large CLR User-Defined Types (UDT) is essentially a `varbinary` type with additional metadata.
See https://msdn.microsoft.com/en-us/library/bb677316.aspx

For example, SQL Server uses `SQL_SS_UDT` type to define its spatial data - exchanged as varbinary.

------

At some point, I will include some test cases for `SQL_SS_UDT` based on SQL Server `GEOMETRY` type (separate PR).

### DIFF
--- a/src/nanodbc.cpp
+++ b/src/nanodbc.cpp
@@ -51,6 +51,16 @@
 #include <sql.h>
 #include <sqlext.h>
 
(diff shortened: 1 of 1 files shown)


lexicalunit:
LGTM

PR #153, opened 2016-05-12
Add empty example for users convenience.

mloskot:
CMake generates project/target with empty C++ source file ready to be filled with user's code to try nanodbc.

-----

The empty example is very convenient and makes it easier for a user to get started with the library.
Modification made to the example do not affect cloned working copy of nanodbc, should not affect any commits.


lexicalunit:
👍 good idea!

PR #154, opened 2016-05-12
Add DB-specific tests for result::affected_rows

mloskot:
Covered for number of DBMS, should serve as documentation of differences in returned value between database/drive implementations.

### DIFF
--- a/test/mssql_test.cpp
+++ b/test/mssql_test.cpp
@@ -22,6 +22,57 @@ namespace
 
 // TODO: catlog_* tests
 
+TEST_CASE_METHOD(mssql_fixture, "affected_rows_test", "[mssql][affected_rows]")
+{
+    // Enable MARS required?
+#if 0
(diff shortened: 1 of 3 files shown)


mloskot:
I'm surprised `affected_rows` test causes SQLite to segfault
https://travis-ci.org/lexicalunit/nanodbc/builds/129878504
```
/home/travis/build/lexicalunit/nanodbc/test/sqlite_test.cpp:42: FAILED:
due to a fatal error condition:
  SIGSEGV - Segmentation violation signal
```
Interestingly, only two of four SQLite build job are failing.

mloskot:
It is unclear where the `SIGSEGV - Segmentation violation signal` comes from, but it looks like the message itself is generated by Catch.

Second, the `sqlite_test` passes on Ubuntu 14.04 without any problems.
On Travis CI's Ubuntu 12.04, it might be failing due some critical issue in that version of the SQLite ODBC Driver.
Or, it might be a bug in Catch.

Fairly annoying.

PR #155, opened 2016-05-13
Add input iterator for result class

mloskot:
The new `result_iterator` should enable C++ idiomatic iteration over rows of query results, also with range-based for loop.

Rename `result::end` to `result::at_end` to avoid name clash with a member `iterator end()` that prevents ADL to work for `nanodbc::begin/end` discovery.

-----

This is a very quickly cooked proposal inspired by question at the end of this https://github.com/lexicalunit/nanodbc/issues/146#issuecomment-217661815. Comments and reviews much appreciated!

### DIFF
--- a/src/nanodbc.cpp
+++ b/src/nanodbc.cpp
@@ -2063,7 +2063,7 @@ public:
         return static_cast<unsigned long>(pos) + rowset_position_;
     }
 
-    bool end() const NANODBC_NOEXCEPT
+    bool at_end() const NANODBC_NOEXCEPT
     {
         if(at_end_)
             return true;
@@ -3850,9 +3850,9 @@ unsigned long result::position() const
     return impl_->position();
 }
 
-bool result::end() const NANODBC_NOEXCEPT
+bool result::at_end() const NANODBC_NOEXCEPT
 {
-    return impl_->end();
+    return impl_->at_end();
 }
 
 bool result::is_null(short column) const
--- a/src/nanodbc.h
+++ b/src/nanodbc.h
@@ -74,6 +74,7 @@
 #ifndef NANODBC_H
 #define NANODBC_H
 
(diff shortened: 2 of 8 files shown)


pepr:
This is nice! In my opinion, switching from the older `.end()` and `.next()` to the C++ classical iteration with `.begin()` and `.end()` would be definitely nice for any C++ developer.

This would also make clear the older issues on how the old `.end()` should behave if `.next()` was not called, yet. One do not have to keep in head two different ways of iteration through the data.

(Not speaking about the range-based loops, that are one of the major "simple" enhancements in newer C++ from programmer's point of view, in my opinion.)

lexicalunit:
Looks good! 👍

mloskot:
@lexicalunit What do you think about renaming `bool end()`?

The proposed `rowset_iterator` can not be used with range-based loops, since ADL won't work due, AFAIU, to existing `bool end()`, so we would have to provide the member `iterator begin()` and `iterator end()` in order to cover all idiomatic iterations.

mloskot:
Finally, I have decided to rename the conflicting `bool end()` to `result::at_end()` to avoid name clash with a member `iterator end()` that prevents ADL to work for `nanodbc::begin/end` discovery.

I also decided there is no need for member `result::begin` and `result::end` functions. In fact, members would introduce extra complexity due to cyclic dependency between `result` and `result_iterator` class.

lexicalunit:
I think it makes sense to rename `end()` to something else since that particular name is typically reserved for iteration in the C++ world. I'm not sure why I called it that in the first place, given that fact. I believe `is_end()` or `at_end()` would be preferable alternatives now that we have an iteration interface, to avoid confusion.

mloskot:
@lexicalunit Right, so do you agree on the current version?

PR #166, opened 2016-05-20
Add test for null value and indicator of variable-length data

mloskot:
This is an attmpt to *evaluate* a correct fix for discovered issue with `SQLFetch` does not setting null indicator for nullable variable-length column.

Related to PR #165.

### DIFF
--- a/test/mssql_test.cpp
+++ b/test/mssql_test.cpp
@@ -237,6 +237,35 @@ TEST_CASE_METHOD(mssql_fixture, "move_test", "[mssql][move]")
     move_test();
 }
 
+TEST_CASE_METHOD(mssql_fixture, "null_access_test", "[mssql][null]")
+{
+    nanodbc::connection connection = connect();
+    drop_table(connection, NANODBC_TEXT("null_access_test"));
+    execute(connection, NANODBC_TEXT("create table null_access_test (a varchar(max));"));
+
+    {
+        nanodbc::statement statement(connection);
+        prepare(statement, NANODBC_TEXT("insert into null_access_test (a) values (?);"));
+        statement.bind_null(0);
+        execute(statement);
+    }
+
+    nanodbc::result results = execute(connection, NANODBC_TEXT("select a from null_access_test;"));
+
+    REQUIRE(results.next());
+    //FIXME:
+    // null indicator is not set for nullable variable-length columns
+    // because auto_bind() for col.blob_ does not specify separate length and indicator buffers
+    // Fix, comment if(col.blob_) branch in auto_bind() to make it call the 2nd SQLBindCol
+    REQUIRE(results.is_null(0));
+    // Since null indicator is not set, empty string is returned instead of the fallback value
+    auto const fallback = NANODBC_TEXT("value_is_null");
+    REQUIRE(results.get<nanodbc::string_type>(0, fallback) == fallback);
+    REQUIRE_THROWS_AS(results.get<nanodbc::string_type>(0).empty(), nanodbc::null_access_error);
+
+    REQUIRE(!results.next());
+}
+
 TEST_CASE_METHOD(mssql_fixture, "null_test", "[mssql][null]")
 {
     null_test();


jon-v:
Have you tested your proposed fix? I don't understand how the second SQLBindCol statement would cause different behavior, given that col.clen_ is 0.

As far as I understand it, the null indicator not being set here is actually a bug in the SQL Server ODBC drivers, which I think we already have the best possible workaround for - see pull request #57.

jon-v:
Some more detail here: http://stackoverflow.com/questions/23063731/get-indicator-value-for-unbound-columns-without-calling-sqlgetdata

mloskot:
@jon-v 
> Have you tested your proposed fix?

Yes, I have tested as below using `Driver={ODBC Driver 11 for SQL Server}`

```
diff --git a/src/nanodbc.cpp b/src/nanodbc.cpp
index fd4e3f5..207def9 100755
--- a/src/nanodbc.cpp
+++ b/src/nanodbc.cpp
@@ -2432,21 +2432,21 @@ private:
         {
             bound_column& col = bound_columns_[i];
             col.cbdata_ = new null_type[rowset_size_];
-            if(col.blob_)
-            {
-                NANODBC_CALL_RC(
-                    SQLBindCol
-                    , rc
-                    , stmt_.native_statement_handle()
-                    , i + 1
-                    , col.ctype_
-                    , 0
-                    , 0
-                    , col.cbdata_);
-                if(!success(rc))
-                    NANODBC_THROW_DATABASE_ERROR(stmt_.native_statement_handle(), SQL_HANDLE_STMT);
-            }
-            else
+            //if(col.blob_)
+            //{
+            //    NANODBC_CALL_RC(
+            //        SQLBindCol
+            //        , rc
+            //        , stmt_.native_statement_handle()
+            //        , i + 1
+            //        , col.ctype_
+            //        , 0
+            //        , 0
+            //        , col.cbdata_);
+            //    if(!success(rc))
+            //        NANODBC_THROW_DATABASE_ERROR(stmt_.native_statement_handle(), SQL_HANDLE_STMT);
+            //}
+            //else
             {
                 col.pdata_ = new char[rowset_size_ * col.clen_];
                 NANODBC_CALL_RC(
diff --git a/test/mssql_test.cpp b/test/mssql_test.cpp
index 2a477f3..bdf3ca8 100644
--- a/test/mssql_test.cpp
+++ b/test/mssql_test.cpp
@@ -171,11 +171,11 @@ TEST_CASE_METHOD(mssql_fixture, "null_access_test", "[mssql][null]")
     // null indicator is not set for nullable variable-length columns
     // because auto_bind() for col.blob_ does not specify separate length and indicator buffers
     // Fix, comment if(col.blob_) branch in auto_bind() to make it call the 2nd SQLBindCol
-    // REQUIRE(results.is_null(0));
+    REQUIRE(results.is_null(0));
     // Since null indicator is not set, empty string is returned instead of the fallback value
-    //auto const fallback = NANODBC_TEXT("value_is_null");
-    // REQUIRE(results.get<nanodbc::string_type>(0, fallback) == fallback);
-    REQUIRE(results.get<nanodbc::string_type>(0).empty());
+    auto const fallback = NANODBC_TEXT("value_is_null");
+    REQUIRE(results.get<nanodbc::string_type>(0, fallback) == fallback);
+    REQUIRE_THROWS_AS(results.get<nanodbc::string_type>(0).empty(), nanodbc::null_access_error);

     REQUIRE(!results.next());
 }
```

> http://stackoverflow.com/questions/23063731/get-indicator-value-for-unbound-columns-without-calling-sqlgetdata

I've come across this SO too, but not being aware of a bug in the new SQL Server driver.

jon-v:
As far as I can tell your fix causes col.clen_ to be 0 and col.pdata_ to point to a new char[0]. So it appears that somehow passing a valid pointer to a zero-length buffer (instead of a null pointer, as the first if branch was doing) causes the null indicator to be correctly written.

If this doesn't break other usage (ie. you can still get non-NULL data from nvarchar(max) columns), then this is great news, and a perfect workaround for the bug.

(The bug is a mismatch with the documentation - quoting from the SO post: "The MS documentation says that it is possible to bind just an indiciator pointer via SQLBindCol and leave dataptr NULL. I've tried this without success. The indiciator gets never modified (initialized with SQL_NULL_DATA). I've tried to set the indicator pointer with SQLSetDescField() manually for the columns (inc desc rec counter, etc.) but nothing changed. It seems that the driver only fills the indicator ptr when a data ptr is available.")

mloskot:
I also suspect the issue is basically with the passing `nullptr`. Possibly, drivers generally expect a typical use case is [indicator buffer and the length buffer are separate buffers](https://msdn.microsoft.com/en-us/library/ms711010.aspx).

Later, I may apply the path pasted above and update this PR, then we will see if MySQL, PostgreSQL, SQLite are happy on Travis CI and AppVeyor (+SQL Server).

jon-v:
Unfortunately it doesn't look like it works - I tested it on my end, and in the non-NULL case the SQLData call that attempts to read the data returns an error ("Invalid Descriptor Index", SQLSTATE 07009).

So I'm still of the opinion that the existing workaround (only returning the correct value from is_null after get/get_ref has been called) is the best possible solution, at least for SQL Server. But it may be possible to dynamically detect a driver's capabilities and ensure proper is_null behavior when using more capable drivers.

mloskot:
>  I tested it on my end, and in the non-NULL case the SQLData call that
> attempts to read the data returns an error ("Invalid Descriptor Index", SQLSTATE 07009).

What database/driver did you test?

jon-v:
SQL Server 2014 / ODBC Driver 11 for SQL Server

mloskot:
On 20 May 2016 at 12:22, jon-v <notifications@github.com> wrote:

> SQL Server 2014 / ODBC Driver 11 for SQL Server
>

Works for me as explained above with SQL Server 2016 / ODBC Driver 11 for
SQL Server

mloskot:
I've just updated this PR disabling the `SQLBindCol` passing `nullptr` to evaluate behaviour against more databases on the CI services.

mloskot:
On Linux, all databases are happy with the changes: [build 451](https://travis-ci.org/lexicalunit/nanodbc/builds/131661694) 

On Windows, [SQL Server 2014 / ODBC Driver 11 for SQL Server](https://ci.appveyor.com/project/lexicalunit/nanodbc/build/1.0.129/job/f6ssbh1h6crsnd9e) failed indeed, whereas [MySQL, PostgreSQL and SQLite are good](https://ci.appveyor.com/project/lexicalunit/nanodbc/build/1.0.129).

The SQL Server case puzzles me is that the same driver has been tried, just different backend.

jon-v:
That is surprising, yes - I would have expected this to be a driver issue, not a backend issue.

As for the other databases, I'd be curious to see how many of them pass this test without the corresponding change in auto_bind.

jon-v:
I'd also be very curious to see what SQLGetInfo returns for SQL_GETDATA_EXTENSIONS on your system (when connecting to SQL Server 2016).

As far as I can tell there are two possible reasons why the tests with this PR fail with 2014 but not with 2016:

1. Something unknown changed with 2016 where passing a zero-length buffer now causes the null indicator to be correctly set without treating the column as bound (which is what is happening with 2014 - passing a real pointer instead of null causes the column to be bound, preventing us from using SQLGetData later, because SQL_GD_BOUND is unsupported).

2. The column is still being considered bound, but 2016 actually supports SQL_GD_BOUND, meaning  SQLGetData can still be called to retrieve the data.

Of these two, the latter seems more likely to me, but I'm unable to find any hint of such a change in release notes or documentation. Unfortunately I don't have access to 2016 at the moment so I can't investigate further myself.

mloskot:
@jon-v 
>  I'd be curious to see how many of them pass this test without the corresponding change in auto_bind.
> I mean the null_access_test.

Good idea. I have split the PR:
* This #166 only add the **null_access_test**.
* New #168 evaluates the change to `auto_bind()`.

I'm going to verify my observations on fresh SQL Server 2016 installation, so I'll be back to your other comments.

mloskot:
@jon-v 
> I'd also be very curious to see what `SQLGetInfo` returns for `SQL_GETDATA_EXTENSIONS`
> on your system (when connecting to SQL Server 2016).

```
auto conn = connect();
SQLUINTEGER gd;
SQLRETURN rc;
rc = SQLGetInfo(conn.native_dbc_handle(), SQL_GETDATA_EXTENSIONS, (SQLPOINTER)&gd, 0, 0);
if (gd & SQL_GD_ANY_ORDER)
    printf("SQL_GD_ANY_ORDER\n");
if (gd & SQL_GD_ANY_COLUMN)
    printf("SQL_GD_ANY_COLUMN\n");
if (gd & SQL_GD_BLOCK)
    printf("SQL_GD_BLOCK\n");
if (gd & SQL_GD_BOUND)
    printf("SQL_GD_BOUND\n");
if (gd & SQL_GD_OUTPUT_PARAMS)
    printf("SQL_GD_OUTPUT_PARAMS\n");
```
outputs:
```
SQL_GD_BLOCK
SQL_GD_OUTPUT_PARAMS
```

I have run against `select @@version`:
```
Microsoft SQL Server 2016 (RC0) - 13.0.1100.286 (X64)   Feb 27 2016 11:45:01
  Copyright (c) Microsoft Corporation
Express Edition (64-bit) on Windows 10 Pro 6.3 <X64> (Build 10586: ) 
```

Regarding my mention about installing fresh SQL Server 2016 environment, I have failed: I couldn't get SQL Server 2016 R3 installed on Vagrant VM running Hyper-V Server 2012 R2.

mloskot:
Finally, it looks @jon-v observation above https://github.com/lexicalunit/nanodbc/pull/166#issuecomment-220567949 has been confirmed for all SQL Servers on AppVeryor: 2008, 2014, 2016 (see also my comment https://github.com/lexicalunit/nanodbc/pull/168#issuecomment-232882393).

I'm closing this PR as non-issue for now.

PR #168, opened 2016-05-20
Verify setting separate value/indicator buffers for all SQLBindCol

mloskot:
This is an extension to #166:

* Also, set separate value/indicator buffers for all `SQLBindCol`, not only for non-BLOB data.
* This is an attmpt to *evaluate* if it is a correct fix for discovered issue with `SQLFetch` does not setting null indicator for nullable variable-length column.

So, we can compare how the new **null_access_test** behaves with and without the fix.

### DIFF
--- a/src/nanodbc.cpp
+++ b/src/nanodbc.cpp
@@ -2645,21 +2645,21 @@ private:
         {
             bound_column& col = bound_columns_[i];
             col.cbdata_ = new null_type[rowset_size_];
-            if(col.blob_)
-            {
-                NANODBC_CALL_RC(
-                    SQLBindCol
-                    , rc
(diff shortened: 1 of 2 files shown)


mloskot:
I have rabased this PR on the latest master.
Fresh build https://ci.appveyor.com/project/lexicalunit/nanodbc/build/1.0.266
seem to confirm @jon-v observations about 
```
[Microsoft][ODBC Driver 11for SQL Server]Invalid Descriptor Index
````
mentioned here https://github.com/lexicalunit/nanodbc/pull/166#issuecomment-220567949

I am going to close both PRs, this #168 and #166 as non-issue.

If new reports about null indicators appear, we will investigate it.

PR #172, opened 2016-05-24
Define own main for Catch

mloskot:
Add support for custom command line option(s).
Add `--connection-string` option to specify connection string for tests (NANODBC_TEST_CONNSTR environment variable still available as fallback).


lexicalunit:
Neat!

PR #175, opened 2016-05-26
Compile on MinGW

dsogari:


mloskot:
Before this is merged, we need to add MinGW build job to our CI setup.

I don't think anyone builds with MinGW regularly.

dsogari:
That's ok. I'll wait.

mloskot:
MinGW on Travis CI is too old to build nanodbc (see #181).
I can't find MinGW for Ubuntu 12.04 with GCC 5.
So, the build is failing https://travis-ci.org/lexicalunit/nanodbc/builds/133210648

dsogari:
Thanks for your effort, @mloskot. I did a search on Ubuntu packages and, as you said, in 12.04LTS there isn't a MinGW with GCC 5.

But it looks like the build job which is failing, fails during CMake configuration steps. And if I understand correctly, it does not have the alterations to CMakeLists.txt that are needed for the MinGW case.

Anyway, if you're willing to accept these modifications, I believe they won't affect other compilers.

mloskot:
Yes it is CMake warning:
```
Compiler /usr/bin/i586-mingw32msvc-g++ has no C++ 0x/11/14 support.
```
and it indicates a compilatio failure is likely going to happen. For instance, libstdc++ prior GCC 5 doesn't support `<codecvt>`.

dsogari:
Ah, I see now. Yes, that would cause failure in compilation.

mloskot:
FYI, We have now MinGW builds on AppVeyor (#196).
I'm going to cherry pick and copy @dsogari 's changes manually to at least make the compilation phase pass.

PR #176, opened 2016-05-26
Fix integer conversions

dsogari:


mloskot:
Is this intentional all the subsequent pul requests that follow #175 include all the commits accumulate? 
IMHO it is not desired practice and each PR should  include unique and preferably single commit related to topic of particular pul request.

dsogari:
You're right, the commits should not be accumulated. I will recreate those PR with only the requested changes. But I need the fixes for MinGW compilation, since it what I use.

As for the DBMS specific details, I don't like them also. However, your suggestion will only work if the requested ODBC C data type corresponds to the largest capacity. I had already tested with SQL_C_LONG and a buffer size of 64 bits, but it does not work. The buffer gets garbage on the last 4 bytes.

mloskot:
You don't have to re-create the PR, just ammend this one:

```
git commit --amend
git push -f dsogari fix-integer-conversion   # force push will replace current PR commit(s)
```

dsogari:
Thanks for the tip using `git commit --amend`.

Yes, you'll have to be careful with this PR.

Meanwhile, I think I'll keep the changes in my forked repository. Thanks again.

mloskot:
@dsogari Can you tell exactly which verion of MinGW you use to compile nanodbc? On Windows or Linux?

dsogari:
It's mingw-w64 on Windows. `gcc --version` gives this output (first line):
> gcc.exe (Rev5, Built by MSYS2 project) 5.3.0

mloskot:
What is the mingw-w64 version (not GCC) exactly? Is it 4.0.5?

dsogari:
I could not find this info anywhere in the installation. My guess is it's 4.0+, because it was installed in 2015. They provide an updater, which I believe would tell the exact version, but its repo URLs are broken. Anyhow, I'll speak with the project owners to sort this out.

Meanwhile, this is the output from `uname -s`:
> MINGW32_NT-10.0

That does not help much

dsogari:
Yes, it's 4.0. I found it [here](https://github.com/msys2/mingw-w64/blob/master/configure.ac), from this repo: https://github.com/msys2/mingw-w64. At least I believe it is this one they use.

mloskot:
Thanks. We may try to install MinGW on AppVeyor instead. For example [drmingw/appveyor.yml](https://github.com/jrfonseca/drmingw/blob/master/appveyor.yml).

dsogari:
That's a good example you found. TBH, I'm not familiar with this CI thing. My guess is that you might have trouble trying to set it up for Windows images.

dsogari:
Not related to this PR, but in your master branch in src/nanodbc.h file, I found the use C++14 automatic return type deduction. In a header, this will force users of the library to switch to the C++14 standard as well. I don't know if this was intended, but I just wanted to let you know.

mloskot:
Yes, AFAIK, it is intentional to use C++14

lexicalunit:
I definitely intend to use he latest standard available when developing nanodbc. It might be good however to make clear what versions expect which standard of c++. Users who are not compiling with the latest standard would be able to use an earlier release before code was added which uses that standard.

mloskot:
I guess, this has become dependant on #184 which should be accepted first.
Then, since this PR mixes changes related to MinGW support and changes in CMakeLists.txt, it needs to be cleared to make the changes exclusively focused on the integer conversion support.

dsogari:
I guess you're right. Since I already tested the changes in my machine, I won't be needing to compile it anymore. Then, I'll make a commit reverting the changes related to MinGW support.

mloskot:
This looks fine. Unless you can provide it, I will try to work on tests to cover this code. Then, @lexicalunit may give green light to merge.

dsogari:
That's great, tnx! Unfortunately, the tests I wrote weren't dedicated to nanodbc. I only checked correctness for the SQLite driver on a private project using nanodbc.

mloskot:
I have managed to confirm the bug locally and proposed test for it in #190.

@lexicalunit I suggest we merge this PR, then merge the #190.

lexicalunit:
👍  Sounds like a plan, @mloskot!

mloskot:
@lexicalunit Sweet, cheers!

dsogari:
@mloskot concerning your comment on the tests, I looked at code of sqlite3odbc.c from [here](http://www.ch-werner.de/sqliteodbc/) and did not found any hardcoded number that SQLDescribeCol uses to obtain the column size. Apparently, the value is retrieved from the database engine.

One way to deal with this lmitation is to apply a hardcoded value for the case of `SQL_C_SBIGINT`in nanodbc.cpp itself. For example:
```
...
std::string buffer;
column_size = 20; // <-- inserted line here
buffer.reserve(column_size + 1); // ensure terminating null
...
```

This implies declaring the local variable `column_size` as non-const:
```
SQLULEN column_size = col.sqlsize_;
```
It will allow the conversion of values such as `9223372036854775807` and `-9223372036854775808`, which you could add to the tests if you're willing to apply this ugly patch.

mloskot:
@dsogari Thanks for checking.
Yup. using large enough size for all integers sounds like a viable idea.
First, I will test the integer conversions with other DBMS' and see how common is that issue.

PR #180, opened 2016-05-26
ANSI/Unicode and static/shared (recreated)

dsogari:
Build all combinations of ANSI/Unicode and static/shared library versions. This is very convenient for package managers such as pacman, because in this case the libraries are pre-compiled and the user has no choice but to install what was compiled.

I believe this will help nanodbc become more widely available both in Linux and Windows (please take a look at the [MINGW-packages](https://github.com/Alexpux/MINGW-packages) of the MSYS2 project, where I already put your library, but had to make some of these patches).

Note that I had to up the CMake minimum required version to 3.0+, in order to allow use of the `target_compile_definitions` directive.


CMakeLists.txt
```diff
@@ -1,11 +1,9 @@
-cmake_minimum_required(VERSION 2.8.7)
+cmake_minimum_required(VERSION 3.0.2)
```
mloskot: We cannot accept this change, since Travis CI is based on Ubuntu 12.04 with CMake **2.8.7**.

dsogari: I see. That's a pity.

dsogari:
I guess I'll close this PR. Anyway, thanks for your time!

CMakeLists.txt
lexicalunit: @dsogari are your other changes not compatible with CMake version 2.8.7? Your PR could be updated to revert this particular change if so.

dsogari: You're right @lexicalunit. I just realized there is a `COMPILE_DEFINITIONS` property for targets in version 2.8.7, which could be used instead of `target_compile_definitions` (from 3.0.2). I'update the PR to reflect this change and some other fixes. Thanks!

dsogari: Should I open a new PR, then? Or is it possible to reopen a closed one?

lexicalunit: @dsogari I believe it is possible to re-open a PR after a force push updates the branch for it. I think the original PR's author has to reopen it. The reopen button is grayed out for me.

dsogari: It's grayed out for me too. The tooltip says "The ... branch was force-pushed or recreated".

mloskot:
FYI, We have now MinGW builds on AppVeyor (#196).
I'm going to cherry pick and copy @dsogari 's changes manually to at least make the compilation phase pass.

PR #182, opened 2016-05-29
Use correct wide char count when copying from SQLGetData buffer

jon-v:
The second parameter to out.append is a character count, not a byte count. Prior to this PR the returned string was twice as long as it should be, the latter half garbage.


mloskot:
Good catch! LGTM

lexicalunit:
Yep, LGTM!

PR #187, opened 2016-06-11
statement_impl::async* members were left uninitialized if not built-in

mloskot:
Addresses new issues reported by the Coverity Scan.
Prefer nullptr instead of NULL.
Clean up trailing whitespaces.


mloskot:
All builds are green and since it's a non-intrusive clean up, I'm merging it.

PR #192, opened 2016-07-01
Add nanodbc::list_drivers free function

mloskot:
Convenient wrapper for ODBC API call `SQLDrivers`.

Add two utility functions to base_test_fixture: `connection_string_parameter` and `iequals_string`,
which are used in `list_drivers` test exercising the new wrapper.

------

Note, if we decide to publish other string utilities like `convert`, that actually could be convenient for nanodbc users, then we may want to publish `iequals_string` as `nanodbc::iequals` as well.

### DIFF
--- a/src/nanodbc.cpp
+++ b/src/nanodbc.cpp
@@ -552,7 +552,7 @@ namespace
     };
 
     // Allocates the native ODBC handles.
-    inline void allocate_handle(SQLHENV& env, SQLHDBC& conn)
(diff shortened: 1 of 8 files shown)


mloskot:
@lexicalunit what do you think, like it? :)

lexicalunit:
Very cool! Looks like solid work to me 👍

mloskot:
Thanks!

PR #193, opened 2016-07-07
Add two catalog operations: list_catalogs and list_schemas

mloskot:
Both functions query data source using ODBC API call SQLTable:
* `list_catalogs` with `SQL_ALL_CATALOGS to get list of valid catalogs.
* `list_schemas` with `SQL_ALL_SCHEMAS` to get list of valid schemas.

------
Important detail to note is that the current `nanodbc::catalog` look-up functions have not been tested 
in the [Identifier Arguments](https://msdn.microsoft.com/en-us/library/ms714579.aspx) mode, so it is (implicitly) assumed the connection/statement attribute `SQL_ATTR_METADATA_ID` is `SQL_FALSE`.

------
There is number of quirks related to use of `SQLTables`, so it is worth to have a look at these pages to get better understanding:
* [Arguments in Catalog Functions](https://msdn.microsoft.com/en-us/library/ms716447.aspx)
* [SQLTables - MariaDB Knowledge Base](https://mariadb.com/kb/en/sql-99/sqltables/)

### DIFF
--- a/src/nanodbc.h
+++ b/src/nanodbc.h
@@ -1309,10 +1309,16 @@ inline auto end(result& /*r*/) { return result_iterator(); }
 //                                                  "Y88P"
 // MARK: Catalog -
 
+//! \brief A resource for get catalog information from connected data source.
+//!
+//! Queries are performed using the Catalog Functions in ODBC.
+//! All provided operations are convenient wrappers around the ODBC API
+//! The original ODBC behaviour should not be affected by any added processing.
 class catalog
 {
 public:
 
+    //! \brief Result set for a list of tables in the data source.
     class tables
     {
     public:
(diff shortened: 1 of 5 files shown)


src/nanodbc.h
```diff
@@ -1469,6 +1469,10 @@ class catalog
       , const string_type& schema = string_type()
       , const string_type& catalog = string_type());
 
+    std::list<string_type> list_catalogs();
```
lexicalunit: No docs?

lexicalunit:
Looks pretty straightforward, just have one comment about missing Doxygen documentation for the new methods. Otherwise 👍

src/nanodbc.h
mloskot: Yea, I rushed too much. I will add some.

mloskot:
Docs updated. Merging.

PR #202, opened 2016-08-17
Add new method result::column_decimal_digits

mloskot:
Add new method result::column_decimal_digits.
Add test for properties of result column descriptor.
Add missing overloads to access column properties by column name.
Re-arrange order of result and result_impl methods.

------
Question regarding the `result::column*` methods:
Some of them are overloaded for both, name and index, (e.g. `column_datatype`) while some are not (e.g. `column_size`). Do we want to unify the interface, provide overloaded variants for all or remove the overloads for name?

### DIFF
--- a/src/nanodbc.cpp
+++ b/src/nanodbc.cpp
@@ -2285,6 +2285,15 @@ public:
         return is_null(column);
     }
 
+    short column(const string_type& column_name) const
+    {
+        typedef std::map<string_type, bound_column*>::const_iterator iter;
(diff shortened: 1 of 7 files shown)


lexicalunit:
Nice! 👍

lexicalunit:
As for the `column_size` and `column_datatype` overloads. My intent was to always provide both overrides in all cases so users could decide if they wanted to refer to columns by name or number.

mloskot:
Cool, I have updated the commit with the missing overloads.

PR #204, opened 2016-08-18
Add catalog::find_table_privileges function

mloskot:
Add `catalog::table_privileges` result set

It is convenient wrapper for ODBC API call `SQLTablePrivileges` and its result set.

### DIFF
--- a/src/nanodbc.h
+++ b/src/nanodbc.h
@@ -1487,8 +1487,6 @@ public:
         /// \note MSDN: This column returns a zero-length string if nullability is unknown.
         ///       ISO rules are followed to determine nullability.
         ///       An ISO SQL-compliant DBMS cannot return an empty string.
-        ///
-        /// TODO: Translate "YES","NO", <empty> strings to IsNullable enum?
         string_type is_nullable() const;
 
     private:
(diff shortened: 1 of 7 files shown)


test/mysql_test.cpp
```diff
@@ -100,6 +100,8 @@ TEST_CASE_METHOD(mysql_fixture, "catalog_tables_test", "[mysql][catalog][tables]
     catalog_tables_test();
 }
 
+// TODO: Add catalog_table_privileges_test - SQLTablePrivileges returns empty result set
```
mloskot: I have tried to call `SQLTablePrivileges`  with variety of arguments (e.g. database name, empty database name, `SQL_NTS` or 0 for size, etc.) and this function always returns nothing. I have no idea why.

I'm using MySQL ODBC 5.3 Driver.

src/nanodbc.h
```diff
@@ -1461,6 +1461,26 @@ class catalog
         result result_;
     };
 
+     //! \brief Result set for a list of tables and the privileges associated with each table.
+    class table_privileges
+    {
+    public:
+        bool next();
```
lexicalunit: If we add `///<` style comments here it will allow Doxygen to pickup these fields in the generated documentation files.

test/mysql_test.cpp
lexicalunit: I'd be ok with noting this issue in the documentation somewhere and omitting this test in MySQL.

lexicalunit:
Looks good! After rebasing on master, running the code style formatter, and add verifying the test pass again you're good to merge 👍

test/mysql_test.cpp
mloskot: After merging, I'll try to look at it again and if still w/o luck, I'll document it.

mloskot:
BTW, I also removed TODO comments: *Translate "YES","NO", <empty> strings to enum*

I think it is better to stick to string values: 1) all catalog values are fetched as strings 2) some (buggy?) drivers may use driver-specific values, then we would be covered.

Tread off: from user's perspective, enums offered by nanodbc would be a convenient simplification of ODBC interface.

PR #210, opened 2016-09-29
Added test for long string on MySQL

saur0n:
Currently nanodbc cuts strings retrieved from MySQL database to first 128 characters.


mloskot:
Please, don't mix unrelated changes in single PR. Please, split them.

PR #214, opened 2016-10-18
Add support for vector of strings input

### LINKED ISSUE #213: Binding multiple string parameters in std::vector
Is it possible to use `bind_strings()` with a `std::vector<const char*>()` object? All of the examples I can find use static arrays only.

If this is not supported (I don't believe it is) could you suggest the best way to copy the data into a contiguous block of memory that can be used?

jimhester:
This makes it easy to do batch insertion of strings.

This is not completely done, I think we would need to add methods to handle null values as well, but I wanted to send it and get some feedback on the implementation, and make sure it passes the CI tests.

Fixes #213

### DIFF
--- a/src/nanodbc.cpp
+++ b/src/nanodbc.cpp
@@ -1690,6 +1690,8 @@ public:
         short param,
         const std::vector<string_type> & values,
         std::size_t elements,
+        const bool* nulls,
+        const string_type::value_type* null_sentry,
         param_direction direction);
 
     // handles multiple null values
@@ -1838,6 +1840,8 @@ void statement::statement_impl::bind_strings(
     short param,
     const std::vector<string_type> & values,
     std::size_t elements,
+    const bool* nulls,
+    const string_type::value_type* null_sentry,
     param_direction direction) {
 
   SQLSMALLINT data_type;
@@ -1852,12 +1856,14 @@ void statement::statement_impl::bind_strings(
       max_len = values[i].length();
     }
   }
+  // add space for null terminator
+  ++max_len;
+
   data_[param] = std::vector<string_type::value_type>(max_len * elements, 0);
   for (std::size_t i = 0; i < elements; ++i) {
-    bind_len_or_null_[param][i] = values[i].length();
     std::copy(values[i].begin(), values[i].end(), data_[param].data() + (i * max_len));
   }
-  bind_parameter(param, data_[param].data(), elements, data_type, param_type, max_len, scale);
+  bind_strings(param, data_[param].data(), max_len, elements, nulls, null_sentry, direction);
 }
 
 void statement::statement_impl::bind_strings(
@@ -1888,11 +1894,11 @@ void statement::statement_impl::bind_strings(
             std::string narrow_rhs;
             narrow_rhs.reserve(s_rhs.size());
             convert(s_rhs, narrow_lhs);
-            if (std::strncmp(narrow_lhs.c_str(), narrow_rhs.c_str(), length))
-                bind_len_or_null_[param][i] = parameter_size;
+            if (!std::strncmp(narrow_lhs.c_str(), narrow_rhs.c_str(), length))
+                bind_len_or_null_[param][i] = SQL_NTS;
 #else
-            if (std::strncmp(s_lhs.c_str(), s_rhs.c_str(), length))
-                bind_len_or_null_[param][i] = parameter_size;
+            if (!std::strncmp(s_lhs.c_str(), s_rhs.c_str(), length))
+                bind_len_or_null_[param][i] = SQL_NTS;
 #endif
         }
     }
@@ -1904,8 +1910,15 @@ void statement::statement_impl::bind_strings(
                 bind_len_or_null_[param][i] = SQL_NTS; // null terminated
         }
     }
+    else
+    {
+      for (std::size_t i = 0; i < elements; ++i)
+      {
+        bind_len_or_null_[param][i] = SQL_NTS;
+      }
+    }
 
-    bind_parameter(param, values, elements, data_type, param_type, parameter_size, scale);
+    bind_parameter(param, values, elements, data_type, param_type, length, scale);
 }
 
 template <>
@@ -3664,7 +3677,7 @@ void statement::bind_strings(
     std::size_t elements,
     param_direction direction) {
 
-  impl_->bind_strings(param, values, elements, direction);
+  impl_->bind_strings(param, values, elements, nullptr, nullptr, direction);
 }
 
 void statement::bind_strings(
@@ -3698,6 +3711,27 @@ void statement::bind_strings(
 {
     impl_->bind_strings(
         param, values, length, elements, nulls, (string_type::value_type*)0, direction);
+}
+
+void statement::bind_strings(
+    short param,
+    const std::vector<string_type> & values,
+    std::size_t elements,
+    const string_type::value_type* null_sentry,
+    param_direction direction)
+{
+    impl_->bind_strings(param, values, elements, (bool*)0, null_sentry, direction);
+}
+
+void statement::bind_strings(
+    short param,
+    const std::vector<string_type> & values,
+    std::size_t elements,
+    const bool* nulls,
+    param_direction direction)
+{
+    impl_->bind_strings(
+        param, values, elements, nulls, (string_type::value_type*)0, direction);
 }
 
 void statement::bind_null(short param, std::size_t elements)
--- a/src/nanodbc.cpp
+++ b/src/nanodbc.cpp
@@ -1143,7 +1143,7 @@ public:
         , open_(false)
         , conn_()
         , bind_len_or_null_()
-        , data_()
+        , string_data_()
 #if defined(NANODBC_DO_ASYNC_IMPL)
         , async_(false)
         , async_enabled_(false)
@@ -1158,7 +1158,7 @@ public:
         , open_(false)
         , conn_()
         , bind_len_or_null_()
-        , data_()
+        , string_data_()
 #if defined(NANODBC_DO_ASYNC_IMPL)
         , async_(false)
         , async_enabled_(false)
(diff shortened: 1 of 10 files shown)


mloskot:
Looks like a very good addition.

The CI build pass, but the new `statement::bind_strings` function does not seem to be covered by tests.

Could you add test case(s) for `statement::bind_strings` ?

src/nanodbc.cpp
```diff
@@ -1823,6 +1835,40 @@ void statement::statement_impl::bind(
...
+            max_len = values[i].length();
+        }
+    }
+    // add space for null terminator
+    ++max_len;
+
+    data_[param] = std::vector<string_type::value_type>(max_len * elements, 0);
+    for (std::size_t i = 0; i < elements; ++i)
+    {
+        std::copy(values[i].begin(), values[i].end(), data_[param].data() + (i * max_len));
```
mloskot: I committed a tweak (c1c0d8a7f1d9c6ffe30e1889fe1b2e5d92510266) that will help to avoid Visual C++ C4996 warning here.

test/base_test_fixture.h
```diff
@@ -1502,6 +1502,38 @@ struct base_test_fixture
         results.get_ref(0, ref);
         REQUIRE(ref == name);
     }
+    void string_vector_test()
```
mloskot: It looks, this test is not called anywhere, is it?

src/nanodbc.cpp
```diff
@@ -1681,6 +1683,15 @@ class statement::statement_impl
         bind(param, values, elements, direction);
     }
 
+    // handles multiple string values
+    void bind_strings(
+        short param,
+        const std::vector<string_type>& values,
+        std::size_t elements,
```
mloskot: Since the vector carries size, the `elements` seem redundant. Would it be possible to get rid of it?

test/base_test_fixture.h
```diff
@@ -1502,6 +1502,38 @@ struct base_test_fixture
...
+        REQUIRE(connection.transactions() == std::size_t(0));
+
+        const std::vector<nanodbc::string_type> first_name = {NANODBC_TEXT("Fred"), NANODBC_TEXT("Barney"), NANODBC_TEXT("Dino")};
+        const std::vector<nanodbc::string_type> last_name = {NANODBC_TEXT("Flintstone"), NANODBC_TEXT("Rubble"), NANODBC_TEXT("")};
+
+        drop_table(connection, NANODBC_TEXT("string_test"));
+        execute(connection, NANODBC_TEXT("create table string_test (first varchar(10), last varchar(10));"));
+
+        nanodbc::statement query(connection);
+        prepare(query, NANODBC_TEXT("insert into string_test(s) values(?, ?)"));
```
mloskot: It looks to me this test is incomplete.

1. It is not called anywhere.
2. SQL statements use incorrect number of columns.
3. `bind_strings` is called with incorrect column : `0` and `2`.
4. Table name `string_test` is used by the other test case, it's better to keep unique table names per test case.

I have taken the liberty to update the tests as follows:

```
diff --git a/test/base_test_fixture.h b/test/base_test_fixture.h
index 311656a..b87df04 100644
--- a/test/base_test_fixture.h
+++ b/test/base_test_fixture.h
@@ -1512,18 +1512,18 @@ struct base_test_fixture
         const std::vector<nanodbc::string_type> first_name = {NANODBC_TEXT("Fred"), NANODBC_TEXT("Barney"), NANODBC_TEXT("Dino")};
         const std::vector<nanodbc::string_type> last_name = {NANODBC_TEXT("Flintstone"), NANODBC_TEXT("Rubble"), NANODBC_TEXT("")};
 
-        drop_table(connection, NANODBC_TEXT("string_test"));
-        execute(connection, NANODBC_TEXT("create table string_test (first varchar(10), last varchar(10));"));
+        drop_table(connection, NANODBC_TEXT("string_vector_test"));
+        execute(connection, NANODBC_TEXT("create table string_vector_test (first varchar(10), last varchar(10));"));
 
         nanodbc::statement query(connection);
-        prepare(query, NANODBC_TEXT("insert into string_test(s) values(?, ?)"));
+        prepare(query, NANODBC_TEXT("insert into string_vector_test(first, last) values(?, ?)"));
         query.bind_strings(0, first_name, 3);
 
         // With null sentry
-        query.bind_strings(2, last_name, 3, NANODBC_TEXT(""));
+        query.bind_strings(1, last_name, 3, NANODBC_TEXT(""));
         nanodbc::execute(query);
 
-        nanodbc::result results = execute(connection, NANODBC_TEXT("select s from string_test;"));
+        nanodbc::result results = execute(connection, NANODBC_TEXT("select first,last from string_vector_test order by first desc;"));
         REQUIRE(results.next());
         REQUIRE(results.get<nanodbc::string_type>(0) == NANODBC_TEXT("Fred"));
         REQUIRE(results.get<nanodbc::string_type>(1) == NANODBC_TEXT("Flintstone"));
diff --git a/test/mssql_test.cpp b/test/mssql_test.cpp
index 12af1a1..2c5d40f 100644
--- a/test/mssql_test.cpp
+++ b/test/mssql_test.cpp
@@ -355,6 +355,11 @@ TEST_CASE_METHOD(mssql_fixture, "string_test", "[mssql][string]")
     string_test();
 }
 
+TEST_CASE_METHOD(mssql_fixture, "string_vector_test", "[mssql][string]")
+{
+    string_vector_test();
+}
+
 TEST_CASE_METHOD(mssql_fixture, "transaction_test", "[mssql][transaction]")
 {
     transaction_test();
```

I tried it against SQL Server 2016 and there seem to be a problem:
* Only first elements from the bound vectors are inserted

```
SELECT [first], [last] FROM [string_vector_test]

first | last
-------------
Fred | Flintstone
```

test/base_test_fixture.h
jimhester: Yes you are right, I did not understand how these tests were being incorporated into each driver test suite when I wrote this, will fix.

mloskot: Yes, there is a bit of fiddling.

Basically,
* The `base_test_fixture.h` is a set of common test cases.
* Each `<database>_test.cpp` is a source code for an independent test program/runner that includes common test cases, plus database-specific cases if such cases exist/are necessary to add.

The procedure is this:
1. Add common test case to `base_test_fixture.h`
2. Copy & paste the `TEST_CASE_METHOD` bolierplate in each of `<database>_test.cpp` files separately, updating name/tags, etc.

If, a feature requires very specific test cases for each database, then skip `base_test_fixture.h` and write dedicated test case in each `<database>_test.cpp` file.

I hope it helps.

test/base_test_fixture.h
```diff
@@ -1502,6 +1502,39 @@ struct base_test_fixture
...
+        nanodbc::statement query(connection);
+        prepare(query, NANODBC_TEXT("insert into string_vector_test(first, last) values(?, ?)"));
+        query.bind_strings(0, first_name);
+
+        // With null sentry
+        query.bind_strings(1, last_name, NANODBC_TEXT(""));
+        nanodbc::execute(query);
+
+        nanodbc::result results = execute(connection, NANODBC_TEXT("select first,last from string_vector_test order by first desc;"));
+        for (auto &vals : results) {
```
mloskot: I don't think this look is right.

Caught by [latest AppVeyor build](https://ci.appveyor.com/project/lexicalunit/nanodbc/build/1.0.321/job/ljosm8j6spte5bji) as ` warning C4189: 'vals': local variable is initialized but not referenced`

test/base_test_fixture.h
jimhester: The issue there was the tests neglected to pass the number of bound elements.

jimhester:
As of https://github.com/lexicalunit/nanodbc/pull/214/commits/8e20009bc08978f9038a3860fb1ea2aa5b33000b all tests now passing (and actually running the test code :))

mloskot:
@jimhester Awesome, thanks! I think it's ready to merge. Let's wait for agreement from @lexicalunit

src/nanodbc.cpp
```diff
@@ -1851,12 +1896,9 @@ void statement::statement_impl::bind_strings(
             std::string narrow_rhs;
             narrow_rhs.reserve(s_rhs.size());
             convert(s_rhs, narrow_lhs);
-            if (std::strncmp(narrow_lhs.c_str(), narrow_rhs.c_str(), length))
```
mloskot: Removal of this `strncmp` leads to [MySQL build job 657.5](https://travis-ci.org/lexicalunit/nanodbc/jobs/170802339) failure on Linux 


```
/home/travis/build/lexicalunit/nanodbc/src/nanodbc.cpp:1900:66: 
    error: cannot convert ‘const char16_t*’ to ‘const char*’ for argument ‘1’ 
         to ‘int strncmp(const char*, const char*, size_t)’
    if (std::strncmp(s_lhs.c_str(), s_rhs.c_str(), length) != 0)
```

mloskot: No problem. That's why CI builds are handy. Thanks!

mloskot:
@jimhester In the master, the current (non-`vector`) binding of strings for batch insertion is broken.

I think I have pinned the bug, see https://github.com/lexicalunit/nanodbc/issues/116#issuecomment-256641921.

Perhaps you could give your feedback - I would highly appreciate it. Also, in case the changes I proposed (described, no PR yet) affect your new `vector`-based implementation.

src/nanodbc.cpp
```diff
@@ -1867,8 +1912,15 @@ void statement::statement_impl::bind_strings(
...
+    else
+    {
+        for (std::size_t i = 0; i < elements; ++i)
+        {
+            bind_len_or_null_[param][i] = SQL_NTS;
+        }
+    }
 
-    bind_parameter(param, values, elements, data_type, param_type, parameter_size, scale);
+    bind_parameter(param, values, elements, data_type, param_type, length, scale);
```
mloskot: See https://github.com/lexicalunit/nanodbc/issues/116#issuecomment-256687931 for detailed discussion related to this change above.

test/base_test_fixture.h
lexicalunit: @mloskot it may be worth adding a section to the README.md describing the process of adding to the test suite. You're outline in this comment would be an excellent addition I think!

PR #215, opened 2016-10-18
get_info(), which gets string information from a connection

jimhester:
This exposes a simple function that takes an info_type returns the information as a string, which allows users to query all available info types (https://msdn.microsoft.com/en-us/library/ms711681(v=vs.85).aspx#Anchor_6) not just those exposed by nanodbc methods.

### DIFF
--- a/src/nanodbc.cpp
+++ b/src/nanodbc.cpp
@@ -883,80 +883,42 @@ public:
 
     void* native_env_handle() const { return env_; }
 
-    string_type dbms_name() const
+    string_type get_string_info(short info_type) const
     {
-        NANODBC_SQLCHAR name[255] = {0};
+        NANODBC_SQLCHAR value[255] = {0};
         SQLSMALLINT length(0);
         RETCODE rc;
         NANODBC_CALL_RC(
             NANODBC_FUNC(SQLGetInfo),
             rc,
             conn_,
-            SQL_DBMS_NAME,
-            name,
-            sizeof(name) / sizeof(NANODBC_SQLCHAR),
+            info_type,
+            value,
+            sizeof(value) / sizeof(NANODBC_SQLCHAR),
             &length);
         if (!success(rc))
             NANODBC_THROW_DATABASE_ERROR(conn_, SQL_HANDLE_DBC);
-        return string_type(&name[0], &name[strarrlen(name)]);
+        return string_type(&value[0], &value[strarrlen(value)]);
+    }
+
+    string_type dbms_name() const
+    {
+      return get_string_info(SQL_DBMS_NAME);
     }
 
     string_type dbms_version() const
     {
-        NANODBC_SQLCHAR version[255] = {0};
-        SQLSMALLINT length(0);
-        RETCODE rc;
-        NANODBC_CALL_RC(
-            NANODBC_FUNC(SQLGetInfo),
-            rc,
-            conn_,
-            SQL_DBMS_VER,
-            version,
-            sizeof(version) / sizeof(NANODBC_SQLCHAR),
-            &length);
-        if (!success(rc))
-            NANODBC_THROW_DATABASE_ERROR(conn_, SQL_HANDLE_DBC);
-        return string_type(&version[0], &version[strarrlen(version)]);
+      return get_string_info(SQL_DBMS_VER);
     }
 
     string_type driver_name() const
     {
-        NANODBC_SQLCHAR name[1024];
-        SQLSMALLINT length;
-        RETCODE rc;
-        NANODBC_CALL_RC(
-            NANODBC_FUNC(SQLGetInfo),
-            rc,
-            conn_,
-            SQL_DRIVER_NAME,
-            name,
-            sizeof(name) / sizeof(NANODBC_SQLCHAR),
-            &length);
-        if (!success(rc))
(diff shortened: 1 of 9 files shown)


src/nanodbc.h
```diff
@@ -1019,6 +1019,10 @@ class connection
     /// \brief Returns the native ODBC environment handle.
     void* native_env_handle() const;
 
+
+    /// \brief Returns information from the ODBC connection as a string.
+    string_type get_string_info(short info_type) const;
```
mloskot: IMO, `get_info` would reflect correspondence to `SQLGetInfo`. The return type indicates the info is returned as a string.

src/nanodbc.cpp
```diff
@@ -881,80 +881,42 @@ class connection::connection_impl
...
-            &length);
-        if (!success(rc))
-            NANODBC_THROW_DATABASE_ERROR(conn_, SQL_HANDLE_DBC);
-        return string_type(&version[0], &version[strarrlen(version)]);
+      return get_string_info(SQL_DBMS_VER);
     }
 
     string_type driver_name() const
     {
-        NANODBC_SQLCHAR name[1024];
```
mloskot: Are we certain 255 is sufficient for all `InfoType` return values?
If not, I'd bump to 1024 or dynamically allocate as per the comment:

```
// FIXME: Allocate buffer of dynamic size as drivers do not agree on universal size
// MySQL driver limits MAX_NAME_LEN=255
// PostgreSQL driver MAX_INFO_STIRNG=128
// MFC CDatabase allocates buffer dynamically.
```

src/nanodbc.cpp
```diff
@@ -881,80 +881,42 @@ class connection::connection_impl
...
-            name,
-            sizeof(name) / sizeof(NANODBC_SQLCHAR),
+            info_type,
+            value,
+            sizeof(value) / sizeof(NANODBC_SQLCHAR),
             &length);
         if (!success(rc))
             NANODBC_THROW_DATABASE_ERROR(conn_, SQL_HANDLE_DBC);
-        return string_type(&name[0], &name[strarrlen(name)]);
+        return string_type(&value[0], &value[strarrlen(value)]);
```
mloskot: Does the new function aim to support all possible `SQLGetInfo` return values?
https://msdn.microsoft.com/en-us/library/ms711681.aspx#Anchor_1
> InfoValuePtr (...) Depending on the InfoType requested, the information returned will be one of the following:
> a null-terminated character string, an SQLUSMALLINT value, an SQLUINTEGER bitmask, an SQLUINTEGER flag,
> a SQLUINTEGER binary value, or a SQLULEN value.

mloskot:
A test for the new function is missing. The aim is to accept new features, even if trivial, only if covered with at least basic test.

src/nanodbc.h
jimhester: That is true, however this function is only intended to return string results, it does no coercion if the results are a non string type, and functions signatures cannot differ solely in their return value.

src/nanodbc.cpp
jimhester: No it is only intended to return null-terminated character strings, we could presumably have corresponding functions for the other types (or at least a `get_integer_info()`) that would handle all of the integer types.

src/nanodbc.h
mloskot: Unless there is a plan to add `get_integer_info` or similar and `get_string_info` is the only function to available, lack of coercion is not relevant. IMHO, `get_string_info` is not a good but ambiguous  name: it does not return an info details about a string (compare with `get_host_info`).  What the function does is: `get_info_as_string`. I may sound anal about naming, but IMHO such details are fairly important for self-descriptive and discoverable interface.

jimhester:
I renamed the function to `get_info()` and made it templated so it now supports all documented return types from `SqlGetInfo`

Also added some simple tests as requested.

mloskot:
@jimhester Thanks for your efforts. The update solution looks good to me. Let's see what @lexicalunit says :)

test/base_test_fixture.h
```diff
@@ -1012,6 +1013,24 @@ struct base_test_fixture
...
+    void get_info_test()
+    {
+        // A generic test to exercise the DBMS info API is callable.
+        // DBMS-specific test (MySQL, SQLite, etc.) may perform extended checks.
+        nanodbc::connection connection = connect();
+        REQUIRE(!connection.get_info<nanodbc::string_type>(SQL_DRIVER_NAME).empty());
+        REQUIRE(!connection.get_info<nanodbc::string_type>(SQL_ODBC_VER).empty());
+
+       // Test SQLUSMALLINT resutls
+        REQUIRE(connection.get_info<unsigned short>(SQL_ODBC_INTERFACE_CONFORMANCE) > 0);
```
mloskot: Querying for `SQL_ODBC_INTERFACE_CONFORMANCE` returns `SQLUINTEGER` (4 byte).

So, `unsigned short` is too narrow type and `SQLGetInfo` filling 4 into 2 bytes `T value` will corrupt the stack.

jimhester: You are right, I was trying to find a value that would return more than 0 for all drivers and somehow missed this was returning a `SQLINTEGER` rather than `SQLUSMALLINT`, will fix.

PR #218, opened 2016-10-26
Add nanodbc::char_type typedef

mloskot:
The `char_type` alias complements `string_type` for convenient access to character type expected by nanodbc interface which depends on build configurations.

---

TODO:
- Shall we also assume `sizeof(NANODBC_SQLCHAR) == sizeof(char_type)` ?

### DIFF
--- a/examples/usage.cpp
+++ b/examples/usage.cpp
@@ -96,7 +96,7 @@ void run_test(nanodbc::string_type const& connection_string)
 
         const size_t elements = 4;
 
-        nanodbc::string_type::value_type xdata[elements][10] = {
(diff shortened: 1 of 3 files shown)


lexicalunit:
Could be useful! I'd be curious if this change helps support a specific real world use case.

mloskot:
My motivation was two fold:
* weaker: when `NANODBC_TEXT` is used, it needs to be assigned to a character type variable - not everyone may wish to use `auto`
* stronger: exact nanodbc-compatible character type is required to define array of strings for batch operations.

mloskot:
Dropping the idea due to lack of general agreement

PR #223, opened 2016-11-10
Do not use auto return types

jimhester:
As far as I can tell this is the only place that nanodbc uses a C++14 feature. As the change is trivial it seems better to keep C++11 compatibility until more C++14 features are used in the future.


lexicalunit:
This seems perfectly reasonable to me. I'll approve and merge if test pass 👍

mloskot:
+1

PR #232, opened 2016-11-25
Allow bindind values of all intrinsic integral types

mloskot:
Allow bindind values of all intrinsic integral types.

Allows direct binding of all intrinsic integral types without intermediate variables defined as `int32_t` or `uint32_t`, etc.

However, C++11 does not guarantee `std::intXY_t` are defined as aliases to intrinsic integral types, this commit assumes the aliases are not indifferent and shifts the priority towards the intrinsic types.

------

Regarding the last note above, C99/C++11 standards references can be found in answers to [Are the fixed width integer types guaranteed to be typedefs for the standard built in types?](http://stackoverflow.com/a/31038099/151641)

### DIFF
--- a/src/nanodbc.cpp
+++ b/src/nanodbc.cpp
@@ -22,6 +22,7 @@
 #include <ctime>
 #include <iomanip>
 #include <map>
+#include <type_traits>
 
 #ifndef __clang__
 #include <cstdint>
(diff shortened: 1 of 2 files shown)


mloskot:
After a few iterations this PR is now ready to review. 

@lexicalunit  What do you think of it?

lexicalunit:
Very cleanly done template metaprogramming. I agree and like this better than the `std::intXY_t` approach. Tests all pass and the code looks correct to me. 🎉

mloskot:
Great, thanks!

PR #234, opened 2016-12-28
Add parameters method to output the number of parameters in a query

jimhester:
Wasn't sure on the best name for this function to follow the current conventions, if you have a better suggestion than `parameters()` I am happy to change it.

I added assertions for parameterized queries throughout the tests rather than writing a completely new test for this, hope that is OK.


lexicalunit:
Yeah given that we already have `parameter_size()` something like `parameters_size()` would just be too confusing. `parameters()` seems good to me. It's less verbose than something like `num_parameters()` or `parameter_count()` or whatever. PR LGTM!

PR #237, opened 2017-01-30
Add result::column_datatype_name method

mloskot:
Calls [SQLColAttribute](https://msdn.microsoft.com/en-us/library/ms713558(v=vs.85).aspx) with the field attribute `SQL_DESC_TYPE_NAME` to obtain the data type name.

The function is complementary to `columns::type_name()`, but invokable on query result to, so it is not necessary to open new [SQLColumns](https://msdn.microsoft.com/en-us/library/ms711683(v=vs.85).aspx) result set.

### DIFF
--- a/src/nanodbc.cpp
+++ b/src/nanodbc.cpp
@@ -2348,6 +2348,37 @@ public:
         return col.sqltype_;
     }
 
+    string_type column_datatype_name(short column) const
+    {
+        if (column >= bound_columns_size_)
+            throw index_range_error();
+
+        NANODBC_SQLCHAR type_name[256] = {0};
+        SQLSMALLINT len = 0; // total number of bytes
+        RETCODE rc;
+        NANODBC_CALL_RC(
+            SQLColAttribute,
+            rc,
+            stmt_.native_statement_handle(),
+            column + 1,
+            SQL_DESC_TYPE_NAME,
+            type_name,
+            sizeof(type_name) / sizeof(NANODBC_SQLCHAR),
+            &len,
+            nullptr);
+        if (!success(rc))
+            NANODBC_THROW_DATABASE_ERROR(stmt_.native_statement_handle(), SQL_HANDLE_STMT);
+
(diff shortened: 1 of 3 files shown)


src/nanodbc.cpp
```diff
@@ -2348,6 +2348,38 @@ class result::result_impl
         return col.sqltype_;
     }
 
+    string_type column_datatype_name(short column) const
```
lexicalunit: I like the function name. Seems straight-forward 👍

src/nanodbc.cpp
```diff
@@ -2348,6 +2348,38 @@ class result::result_impl
...
+
+        NANODBC_ASSERT(len % sizeof(NANODBC_SQLCHAR) == 0);
+        len = len / sizeof(NANODBC_SQLCHAR);
+        return string_type(type_name, type_name + len);
+    }
+
+    string_type column_datatype_name(const string_type& column_name) const
+    {
+        const short column = this->column(column_name);
+        return column_datatype_name(column);
```
lexicalunit: I suppose these two lines could be made into one line but I don't think it matters either way.

src/nanodbc.h
```diff
@@ -1358,6 +1358,24 @@ class result
     /// \brief Returns a identifying integer value representing the SQL type of this column by name.
     int column_datatype(const string_type& column_name) const;
 
+    /// \brief Returns data source�dependent data type name of this column.
+    ///
+    /// The function calls SQLCoLAttribute with the field attribute SQL_DESC_TYPE_NAME to
+    /// obtain the data type name.
+    /// If the type is unknown, an empty string is returned.
+    /// \note Unlike other column metadata functions (eg. column_datatype()),
```
lexicalunit: Good note! Great documentation here.

src/nanodbc.cpp
mloskot: Good point. I've simplified it now.

src/nanodbc.h
mloskot: Thanks!

PR #238, opened 2017-01-30
Fix binding of SQL_DECIMAL/SQL_NUMERIC type as character data

mloskot:
Binding SQL fixed precision and scale data to C float-point value may lead to unexpected rounding or truncation errors, especially during retrieval of min or max values of particular SQL data types.

For example:
Max value of SQL Server data type [MONEY](https://msdn.microsoft.com/en-us/library/ms179882.aspx) is `922337203685477.5807`.
If the `MONEY` column is bound as `SQL_C_DOUBLE`, then the fetched value is rounded to an odd value of `922337203685477.6250`.

------

I have tested the fix against SQL Server only with the relevant test included.

### DIFF
--- a/src/nanodbc.cpp
+++ b/src/nanodbc.cpp
@@ -2627,11 +2627,18 @@ private:
                 break;
             case SQL_DOUBLE:
             case SQL_FLOAT:
-            case SQL_DECIMAL:
             case SQL_REAL:
-            case SQL_NUMERIC:
                 col.ctype_ = SQL_C_DOUBLE;
                 col.clen_ = sizeof(double);
+                break;
+            case SQL_DECIMAL:
+            case SQL_NUMERIC:
+                col.ctype_ = SQL_C_CHAR;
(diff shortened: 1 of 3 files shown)


src/nanodbc.cpp
```diff
@@ -2627,12 +2627,19 @@ class result::result_impl
...
                 col.clen_ = sizeof(double);
                 break;
+            case SQL_DECIMAL:
+            case SQL_NUMERIC:
+                col.ctype_ = SQL_C_CHAR;
+                // SQL column size defines number of digits without the decimal mark
+                // and without minus sign which may also occur.
+                // We need to adjust buffer length allow space for null-termination character
+                // as well as the fractional part separator and the minus sign.
+                col.clen_ = (col.sqlsize_ + 1 + 1 + 1) * sizeof(SQLCHAR);
```
lexicalunit: 👍

mloskot:
Fixing CI build failures...

PR #256, opened 2017-04-05
Rename src directory to nanodbc

mloskot:
Switch to buried header and `#include <nanodbc/nanodbc.h>`

Rename directories with singular names.

Rename scripts to utility.


example/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
-add_custom_target(examples DEPENDS examples)
+add_custom_target(examples DEPENDS examplesae)
```
lexicalunit: Should this be `example` instead of `examplesae`?

mloskot: Thank you for catching this. I was fiddling with the custom target name and `DEPENDS` spec, apparently, missing to restore things to normal.

I'm not clear if `DEPENDS` is needed at all though. Thoughts?

lexicalunit: I believe it allows `make examples` to detect that it needs to be re-compiled if source in the `example/` directory changes.

mloskot: Right, but how it works since this `CMakeLists.txt` is inside the `example/` directory itself. 

The docs say nothing about directories, just single files. How it actually works remains unclear to me. It doesn't harm though.

https://cmake.org/cmake/help/latest/command/add_custom_target.html
> `DEPENDS`
> 
>     Reference files and outputs of custom commands created with add_custom_command()
>     command calls in the same directory (CMakeLists.txt file).

PR #267, opened 2017-05-09
Add test for std::vector<bool> workaround

### LINKED ISSUE #266: Binding null values when using array binding
I am trying to load data into SQL Server using array binding where some values could be null.
For strings I use this overload
```
void bind_strings(
        short param_index,
        std::vector<string_type> const& values,
        string_type::value_type const* null_sentry,
        param_direction direction = PARAM_IN);
```
This kind of works but it inserts a empty string instead of an actual database null
So I thought I should use this overload instead:
```
void bind_strings(
        short param_index,
        std::vector<string_type> const& values,
        bool const* nulls,
        param_direction direction = PARAM_IN); 
```
I guess the 3rd argument  pointer is an array of bools , correct?  What is the correct usage of this overload? I am not having success with it or may be there is a better alternative overload that I should use?

jimhester:
Fixes #266

### DIFF
--- a/test/test_case_fixture.h
+++ b/test/test_case_fixture.h
@@ -1429,6 +1429,45 @@ struct test_case_fixture : public base_test_fixture
         REQUIRE(results.is_null(2));
     }
 
+    void test_string_vector_null_vector()
+    {
+        nanodbc::connection connection = connect();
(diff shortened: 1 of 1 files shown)


mloskot:
Great. Thank you @jimhester !

PR #271, opened 2017-05-11
[clang-format] Run clang-format 3.9+

mloskot:
Reformatting generated with clang-format 3.9 on Windows.
I also tried clang-format 5.0.0 and the result is the same.

@lexicalunit Are you getting the same results? Could you confirm this reformat is correct according to our `.clang-format`?

I wonder reformat is fine or the changes are are due to differences between clang-format 3.8, 3.9 and 5.0

### DIFF
--- a/example/northwind.cpp
+++ b/example/northwind.cpp
@@ -22,8 +22,8 @@ int main()
         for (int i = 1; row.next(); ++i)
         {
             cout << i << " :" << convert(row.get<nanodbc::string>(0)) << " "
-                 << convert(row.get<nanodbc::string>(1)) << " " << convert(row.get<nanodbc::string>(2))
-                 << " " << endl;
+                 << convert(row.get<nanodbc::string>(1)) << " "
+                 << convert(row.get<nanodbc::string>(2)) << " " << endl;
         }
     }
     catch (runtime_error const& e)
--- a/example/usage.cpp
+++ b/example/usage.cpp
@@ -146,10 +146,10 @@ void run_test(nanodbc::string const& connection_string)
         nanodbc::string::value_type const* b_null = NANODBC_TEXT("");
         int a_data[elements] = {0, 88, 0, 0, 0};
         nanodbc::string::value_type b_data[elements][10] = {NANODBC_TEXT(""),
-                                                                 NANODBC_TEXT("non-null"),
-                                                                 NANODBC_TEXT(""),
-                                                                 NANODBC_TEXT(""),
-                                                                 NANODBC_TEXT("")};
+                                                            NANODBC_TEXT("non-null"),
+                                                            NANODBC_TEXT(""),
+                                                            NANODBC_TEXT(""),
+                                                            NANODBC_TEXT("")};
 
         statement.bind(0, a_data, elements, &a_null);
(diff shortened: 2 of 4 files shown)


lexicalunit:
Using `clang-format` version 5.0.0 on my macOS machine on this branch resulted in this diff:

```diff
diff --git a/example/northwind.cpp b/example/northwind.cpp
index 7d00818..52a0b00 100644
--- a/example/northwind.cpp
+++ b/example/northwind.cpp
@@ -22,8 +22,8 @@ int main()
         for (int i = 1; row.next(); ++i)
         {
             cout << i << " :" << convert(row.get<nanodbc::string>(0)) << " "
-                 << convert(row.get<nanodbc::string>(1)) << " " << convert(row.get<nanodbc::string>(2))
-                 << " " << endl;
+                 << convert(row.get<nanodbc::string>(1)) << " "
+                 << convert(row.get<nanodbc::string>(2)) << " " << endl;
         }
     }
     catch (runtime_error const& e)
diff --git a/example/usage.cpp b/example/usage.cpp
index bc52101..4b3bc0f 100644
--- a/example/usage.cpp
+++ b/example/usage.cpp
@@ -146,10 +146,10 @@ void run_test(nanodbc::string const& connection_string)
         nanodbc::string::value_type const* b_null = NANODBC_TEXT("");
         int a_data[elements] = {0, 88, 0, 0, 0};
         nanodbc::string::value_type b_data[elements][10] = {NANODBC_TEXT(""),
-                                                                 NANODBC_TEXT("non-null"),
-                                                                 NANODBC_TEXT(""),
-                                                                 NANODBC_TEXT(""),
-                                                                 NANODBC_TEXT("")};
+                                                            NANODBC_TEXT("non-null"),
+                                                            NANODBC_TEXT(""),
+                                                            NANODBC_TEXT(""),
+                                                            NANODBC_TEXT("")};

         statement.bind(0, a_data, elements, &a_null);
         statement.bind_strings(1, b_data, b_null);
```

lexicalunit:
Based on our `.clang-format` the max `ColumnLimit` is 100. So the line:

```
                 << convert(row.get<nanodbc::string>(1)) << " " << convert(row.get<nanodbc::string>(2))
```

Definitely does seem like it needs to be broken up into multiple lines.

mloskot:
Hmm, looks like there might be clang-format differences. I'll check on Linux later.

mloskot:
I've just run `clang-format 4.0.0` on Linux and I'm getting exactly the same results as you above.

Conclusion: clang-format 4 or later is best bet since version 3.9 or older may give different output than expected.

mloskot:
@lexicalunit I've got a bit confused, since this PR only reformats `nanodbc/` files.

Can you confirm that `clang-format` 5.0 gives you no different results for `nanodbc.cpp` and `nanodbc.h`?

lexicalunit:
Ah, I usually just run `utility/style.sh` so that's why I got the changes in `example/`. I can confirm that I get no differences in the files in `nanodbc/` 👍

mloskot:
I see. OK, I've run `utility/style.sh` and the `example/` files have been updated. Looks good

PR #272, opened 2017-05-13
Got rid of boost::locale dependency on GCC 4

saur0n:
This commit makes using boost for UTF-8 → UTF-16 and UTF-16 → UTF-8 conversion no longer needed. Standard C functions are used to do this.
Links: #59, #44.

### DIFF
--- a/nanodbc/nanodbc.cpp
+++ b/nanodbc/nanodbc.cpp
@@ -18,6 +18,7 @@
 #include <algorithm>
 #include <clocale>
 #include <cstdio>
+#include <cstdlib>
 #include <cstring>
 #include <ctime>
 #include <iomanip>
--- a/nanodbc/nanodbc.cpp
+++ b/nanodbc/nanodbc.cpp
@@ -36,12 +36,12 @@
 #endif
 
 #ifdef NANODBC_ENABLE_BOOST
-#   include <boost/locale/encoding_utf.hpp>
+#include <boost/locale/encoding_utf.hpp>
 #elif defined(__GNUC__) && __GNUC__ < 5
-#   define NANODBC_USE_STDLIB
-#   include <cwchar>
+#define NANODBC_USE_STDLIB
+#include <cwchar>
 #else
-#   include <codecvt>
+#include <codecvt>
 #endif
 
 #ifdef __APPLE__
--- a/nanodbc/nanodbc.cpp
+++ b/nanodbc/nanodbc.cpp
@@ -263,9 +263,7 @@ inline void convert(const wide_string& in, std::string& out)
     using boost::locale::conv::utf_to_utf;
     out = utf_to_utf<char>(in.c_str(), in.c_str() + in.size());
 #elif defined(__GNUC__) && (__GNUC__ < 5)
-    std::vector<wchar_t> characters(in.length());
-    for (size_t i=0; i<in.length(); i++)
-        characters[i] = in[i];
+    std::vector<wchar_t> characters(in.begin(), in.end());
     const wchar_t * source = characters.data();
     size_t size = wcsnrtombs(nullptr, &source, characters.size(), 0, nullptr);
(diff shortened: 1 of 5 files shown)


nanodbc/nanodbc.cpp
```diff
@@ -35,9 +35,12 @@
 #endif
 
 #ifdef NANODBC_ENABLE_BOOST
-#include <boost/locale/encoding_utf.hpp>
+#   include <boost/locale/encoding_utf.hpp>
```
mloskot: Please, avoid indentation here

nanodbc/nanodbc.cpp
```diff
@@ -35,9 +35,12 @@
...
+#elif defined(__GNUC__) && __GNUC__ < 5
+#   define NANODBC_USE_STDLIB
+#   include <cwchar>
 #else
-#include <codecvt>
+#   include <codecvt>
```
mloskot: Please, avoid indentation here too

nanodbc/nanodbc.cpp
```diff
@@ -35,9 +35,12 @@
 #endif
 
 #ifdef NANODBC_ENABLE_BOOST
-#include <boost/locale/encoding_utf.hpp>
+#   include <boost/locale/encoding_utf.hpp>
+#elif defined(__GNUC__) && __GNUC__ < 5
+#   define NANODBC_USE_STDLIB
```
mloskot: If this is GCC-specific feature, should rather read `NANODBC_GCC_USE_STDLIB`

nanodbc/nanodbc.cpp
```diff
@@ -259,17 +262,25 @@ inline void convert(const wide_string& in, std::string& out)
 #ifdef NANODBC_ENABLE_BOOST
     using boost::locale::conv::utf_to_utf;
     out = utf_to_utf<char>(in.c_str(), in.c_str() + in.size());
-#else
-// Workaround for confirmed bug in VS2015 and VS2017 too
-// See: https://connect.microsoft.com/VisualStudio/Feedback/Details/1403302
-#if defined(_MSC_VER) && (_MSC_VER >= 1900)
+#elif defined(NANODBC_USE_STDLIB)
+    std::vector<wchar_t> characters(in.length());
+    for (size_t i=0; i<in.length(); i++)
```
mloskot: * Code formatting is inconsistent with `.clang-format`
* What's the reason behind hand-rolling the loop?

saur0n:
`NANODBC_ENABLE_BOOST` switch and boost conversion routines are not needed now and can be safely removed. What do you think about this?

mloskot:
> What do you think about this?

* https://github.com/lexicalunit/nanodbc/pull/272#pullrequestreview-37977012
* `if defined(__GNUC__) && (__GNUC__ < 5)` - what if I use GCC 6?
* What if user prefers to use `Boost.Locale`? I think there's no reason to remove the optional support of Boost-based conversion.

saur0n:
OK, I agree that boost::locale should not be removed.

lexicalunit:
It would be nice to finally simplify the code and reduce dependencies.

mloskot:
@saur0n Can you address my other questions about (ie. about GCC 4)?

saur0n:
I have not found any way to check for specific C++ features in GCC except checking GCC version.

mloskot:
LGTM, merging this one

@lexicalunit Do you prefer to completely wipe out Boost uses?

mloskot:
@saur0n Thanks!

lexicalunit:
@mloskot yeah it'd be great to reduce the number of dependencies as much as possible :)

saur0n:
OK, I will remove boost dependencies.

mloskot:
@lexicalunit Sounds good,
@saur0n Thanks again.

PR #275, opened 2017-05-30
SAP/sybase ASE odbc driver doesn't set sqlsize to 0 when retrieving varchar columns

ChrisBFX:
The SAP Adaptive Server Enterprise odbc driver seems not to touch the sqlsize parameter at all, when calling SQLDescribeCol on a varchar column. Which results in undefined behavior as the variable is never initialized (this only results in a problem when compiled with -O2 for me, so that was fun to debug).


lexicalunit:
Good catch!

PR #283, opened 2017-09-10
Fix handful of cppcheck warnings

mloskot:
Command:
```
  cppcheck --template "{file}({line}): {severity} ({id}): {message}" \
    --enable=style --force --std=c++11 -j 8 -I nanodbc . 2> cppcheck.txt
```

Result:
* Class 'result_iterator' has a constructor with 1 argument that is not explicit.
* Class 'catalog' has a constructor with 1 argument that is not explicit.
* Struct 'is_iequal' has a constructor with 1 argument that is not explicit.

### DIFF
--- a/nanodbc/nanodbc.h
+++ b/nanodbc/nanodbc.h
@@ -1424,7 +1424,7 @@ public:
     result_iterator() = default;
 
     /// Create result iterator for a given result set.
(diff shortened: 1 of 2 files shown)


lexicalunit:
LGTM!

PR #288, opened 2017-09-18
Adjust parameter size for SQLBindParameter to avoid truncation

### LINKED ISSUE #287: Large Objects: String data, right truncation
## Environment

* nanodbc version: master
* DBMS name/version: SQL Server (w/ 2014, 2016, 2017)
* OS and Compiler: Windows 10 + VS2017

## Actual behavior

Attempt to execute prepared statement with bound large object (`varchar`, `varbinary`, UDT) with size larger than a database/driver-specific limit fails due to truncation.
That is despite schema of table column in database is defined without such limit, becuase nanodbc does not specify "unlimited" size input.

For example, SQL Server has 8K limit, https://docs.microsoft.com/en-us/sql/relational-databases/native-client/features/using-large-value-types:
> Large value data types can have a maximum size between 1 and 8 KB, or they can be specified as unlimited.

So, inserting (bound) object larger than 8K will fail due to trancation.

The issue is that parameter size reported by `SQLDescribeParam` for Large Objects is at at most the upper bound of size limit.:

* For SQL VARBINARY(MAX), it is Zero which actually means SQL_SS_LENGTH_UNLIMITED.
* For SQL UDT (eg. GEOMETRY), it may be driver-specific max limit (eg. SQL Server is `DBMAXCHAR=8000` bytes defined in `msodbcsql.h` of Microsoft ODBC Driver for SQL Server).

## Expected  behavior

Adjust parameter size for `SQLBindParameter` to avoid truncation, according to this table at https://docs.microsoft.com/en-us/sql/relational-databases/native-client/odbc/large-clr-user-defined-types-odbc:

sqlserver-size

## Minimal Working Example

```
connection c1("...");
statement s1(c1);
prepare(s1, INSERT INTO t (g) VALUES (geometry::STGeomFromWKB(?, 0)));

// blob is binary object with size larger than 8K (SQL Server limit)
std::vector<std::vector<std::uint8_t>> rows = {blob};

s1.bind(0, rows);
execute(s1);
// 22001: [Microsoft][ODBC Driver 11for SQL Server] String data, right truncation
```

mloskot:
Ensure large object is bound either with size less than parameter size reported by `SQLDescribeParam` or with `SQL_SS_LENGTH_UNLIMITED`.
This is to avoid error `SQLSTATE=22001: String data, right-truncated`

Fixes #287


mloskot:
Travis CI is experiencing lags (hmm, OSX builds), perhaps it's related to the service issue
https://twitter.com/traviscistatus/status/909845838721163264

mloskot:
Merged in https://github.com/nanodbc/nanodbc/commit/4d7a1ef81314c60d3a05274920d3b12ddc863b33

PR #297, opened 2017-10-10
 Allow parameter binding and result::get of narrow string when Unicode support is enabled

ChrisBFX:
I guess I have to explain this one a bit:
We are using utf8 in our database (or at least we assume utf8) but don't use the unicode api, which works somewhat fine as we are the only one writing/reading. The problem we encountered then was that the error messages and connection parameters (username password) are encoded in whatever the system uses as a default. So we enabled the nanodbc unicode support and while this solves the encoding problem we would have to change our internal data structures to be able to bind parameters and using get_ref (basically those which take a string by non-const reference). Every other function call is fine for us, because we can simply convert the arguments/return values.
I think it would make sense to always support string and wide_string for data types, regardless of the odbc-api used, the same way it is ok to use 64bit integers on a 32bit system.

So this pr makes the following changes:
- nanodbc::wide_string is now always defined in the header
- new type traits, to test if a type is a string or wide_string, char or wide_char_t
- statement::bind_strings now always allows string and wide_string as paremeter (this is realized with SFINAE and the new type traits), I know it is long planned to simplify the bind api, and with this it gets even more complex (sorry about that ;)) but I wanted to stay backwards compatible 
- bind allows string and wide_string
- specializations of result::get for string and wide_string
- convert from string to wide_string is now always possible (I hope this works for everyone)

There were also a bug in the convert function and in the the test fixtures under gcc. (If needed I can do another pull request only with these changes).

### DIFF
--- a/nanodbc/nanodbc.cpp
+++ b/nanodbc/nanodbc.cpp
@@ -116,6 +116,10 @@
 // MARK: Unicode -
 // clang-format on
 
+// Import string types defined in header file, so we don't have to type nanodbc:: everywhere
+using nanodbc::wide_char_t;
+using nanodbc::wide_string;
+
 #ifdef NANODBC_ENABLE_UNICODE
 #define NANODBC_FUNC(f) f##W
 #define NANODBC_SQLCHAR SQLWCHAR
@@ -125,18 +129,14 @@
 #endif
 
 #ifdef NANODBC_USE_IODBC_WIDE_STRINGS
-typedef std::u32string wide_string;
 #define NANODBC_CODECVT_TYPE std::codecvt_utf8
 #else
 #ifdef _MSC_VER
-typedef std::wstring wide_string;
 #define NANODBC_CODECVT_TYPE std::codecvt_utf8_utf16
 #else
-typedef std::u16string wide_string;
 #define NANODBC_CODECVT_TYPE std::codecvt_utf8_utf16
 #endif
 #endif
-typedef wide_string::value_type wide_char_t;
 
 #if defined(_MSC_VER)
 #ifndef NANODBC_ENABLE_UNICODE
@@ -264,7 +264,7 @@ inline void convert(const wide_string& in, std::string& out)
     out = utf_to_utf<char>(in.c_str(), in.c_str() + in.size());
 #elif defined(__GNUC__) && (__GNUC__ < 5)
     std::vector<wchar_t> characters(in.begin(), in.end());
-    const wchar_t * source = characters.data();
+    const wchar_t* source = characters.data();
     size_t size = wcsnrtombs(nullptr, &source, characters.size(), 0, nullptr);
     if (size == std::string::npos)
         throw std::range_error("UTF-16 -> UTF-8 conversion error");
@@ -281,7 +281,6 @@ inline void convert(const wide_string& in, std::string& out)
 #endif
 }
 
-#ifdef NANODBC_ENABLE_UNICODE
 inline void convert(const std::string& in, wide_string& out)
 {
 #ifdef NANODBC_ENABLE_BOOST
@@ -292,7 +291,7 @@ inline void convert(const std::string& in, wide_string& out)
     if (size == std::string::npos)
         throw std::range_error("UTF-8 -> UTF-16 conversion error");
     std::vector<wchar_t> characters(size);
-    const char * source = in.data();
+    const char* source = in.data();
     mbsnrtowcs(&characters[0], &source, in.length(), characters.size(), nullptr);
     out = std::string(characters.begin(), characters.end());
 #elif defined(_MSC_VER) && (_MSC_VER >= 1900)
@@ -311,12 +310,11 @@ inline void convert(const wide_string& in, wide_string& out)
 {
     out = in;
 }
-#else
+
 inline void convert(const std::string& in, std::string& out)
 {
     out = in;
 }
-#endif
 
 // Attempts to get the most recent ODBC error as a string.
 // Always returns std::string, even in unicode mode.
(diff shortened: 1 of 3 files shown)


nanodbc/nanodbc.h
```diff
@@ -157,6 +157,18 @@ typedef std::string string;
...
+typedef std::u32string wide_string;
+#else
+#ifdef _MSC_VER
+typedef std::wstring wide_string;
+#else
+typedef std::u16string wide_string;
+#endif
+#endif
+
+typedef wide_string::value_type wide_char_t;
```
mloskot: 1. Any particular reason to not to rename to `nanodbc::wchar_t`  and `nanodbc::wstring`?

2. So, if built with `NANODBC_ENABLE_UNICODE` then `nanodbc::string` and `nanodbc::wide_string` alias the same type (eg. `std::wstring`), but if Unicode is not enabled, then they are distinct string types. Is that the intention?

nanodbc/nanodbc.cpp
```diff
@@ -2062,12 +2055,31 @@ void statement::statement_impl::bind_strings(
...
 template <>
+bool statement::statement_impl::equals(const std::string& lhs, const std::string& rhs)
+{
+    return std::strncmp(lhs.c_str(), rhs.c_str(), lhs.size()) == 0;
+}
+
+template <>
+bool statement::statement_impl::equals(const wide_string& lhs, const wide_string& rhs)
+{
+    // convert narrow strings, cause wcsncmp doesn't work on some machines
```
mloskot: machines? You mean compilation toolkits? Any example?

nanodbc/nanodbc.cpp
```diff
@@ -3357,9 +3390,7 @@ connection::connection(const connection& rhs)
 }
 
 #ifndef NANODBC_NO_MOVE_CTOR
-connection::connection(connection&& rhs) NANODBC_NOEXCEPT : impl_(std::move(rhs.impl_))
-{
-}
+connection::connection(connection&& rhs) NANODBC_NOEXCEPT : impl_(std::move(rhs.impl_)) {}
```
mloskot: Unless you did run clang-format, this looks like an accidental change and should not be mixed in.

nanodbc/nanodbc.cpp
```diff
@@ -2936,7 +2968,8 @@ inline void result::result_impl::get_ref_impl<string>(short column, string& resu
             // Type is unicode in the database, convert if necessary
             const SQLWCHAR* s =
                 reinterpret_cast<SQLWCHAR*>(col.pdata_ + rowset_position_ * col.clen_);
-            const string::size_type str_size = col.cbdata_[static_cast<size_t>(rowset_position_)] / sizeof(SQLWCHAR);
+            const string::size_type str_size =
```
mloskot: Unless you did run clang-format, this looks like an accidental change and should not be mixed in.

mloskot:
I think this is useful to support character data exchange via narrow and wide character strings, regardless if Unicode is enabled.

I'm not sure how to document role of `NANODBC_ENABLE_UNICODE` now to make it clear. It used to control
1. [Unicode-ness of ODBC API](https://docs.microsoft.com/en-us/sql/odbc/reference/develop-app/unicode-function-arguments) (`fooA` vs `fooW`)
2. `nanodbc::string` type used in the library interface

With this PR, it still does the 1. but not really, or partially, the 2.

nanodbc/nanodbc.cpp
ChrisBFX: I've looked at the git log why not simply wcsncmp was used and the related commit message was:
e6059ff3a79062f83256b9d1d3c9c8368798781e
> Functions like `swprintf()`, `wcsftime()`, `wcsncmp()` can not be used
> with `u16string` types. Instead, prefers to narrow unicode string to
> work with them, and then widen them after work has been completed.

I will change the code comment to this so future readers will understand it better.

nanodbc/nanodbc.cpp
ChrisBFX: Yes I've run clang-format. Should i revert these changes?

nanodbc/nanodbc.cpp
mloskot: Thanks for looking at this. Good idea about the comment.

nanodbc/nanodbc.cpp
mloskot: Oh, I see. Then possibly I it's my bad and I had forgotten to run clang-format before a commit.

>  Should i revert these changes?

Best would be if style changes are submitted as separate commits/PRs.

If it would not be a problem for you to revert the style changes, then I could run clang-format, commit, then you could rebase your `ChrisBFX:string_refactor` to update this PR.

If reverting the style changes is a trouble, we could swallow it this time, unless @lexicalunit objects.

lexicalunit: I'm ok with style changes being applied with PRs that add functionality sometimes. It's not ideal but it's not the end of the world. We shouldn't make it a habit if we can avoid it easily.

lexicalunit:
We could break style changes that aren't part of this refactor to their own commit and keep them as part of this PR if it's not too much trouble. Pretty easy to do with git's interactive rebase. If it's too much trouble though I'd rather accept the PR than let it linger and grow stale.

nanodbc/nanodbc.h
ChrisBFX: 1. I would prefer not to use the standard names, as this could result in some strange errors when the nanodbc:: namespace is mistakenly omitted.
2. Yes thats basically it.

ChrisBFX:
I've removed the style changes from the first commit.

mloskot:
@ChrisBFX Thank you

PR #298, opened 2017-10-11
NUMERIC and DECIMAL can now be retrieved using arithmetic types again

ChrisBFX:
In #238 the behavior of NUMERIC and DECIMAL was changed, so that nanodbc uses strings internally to retain precision. Which is great, but unfortunately the change was done such that a `get<int64_t>()` on a NUMERIC now casts the first char of the string representation to an int.

This PR tries to convert the string to wanted type using stringstream.

### DIFF
--- a/nanodbc/nanodbc.cpp
+++ b/nanodbc/nanodbc.cpp
@@ -2528,11 +2528,17 @@ public:
     }
 
 private:
-    template <class T>
+    template <class T, typename std::enable_if<!is_string<T>::value, int>::type = 0>
     void get_ref_impl(short column, T& result) const;
 
     template <class T, typename std::enable_if<is_string<T>::value, int>::type = 0>
-    void get_ref_impl(short column, std::basic_string<typename T::value_type>& result) const;
+    void get_ref_impl(short column, T& result) const;
+
+    template <class T, typename std::enable_if<!is_character<T>::value, int>::type = 0>
+    void get_ref_from_string_column(short column, T& result) const;
+
+    template <class T, typename std::enable_if<is_character<T>::value, int>::type = 0>
+    void get_ref_from_string_column(short column, T& result) const;
 
     void before_move() NANODBC_NOEXCEPT
     {
@@ -2852,9 +2858,7 @@ inline void result::result_impl::get_ref_impl<timestamp>(short column, timestamp
 }
 
 template <class T, typename std::enable_if<is_string<T>::value, int>::type>
-inline void result::result_impl::get_ref_impl(
-    short column,
-    std::basic_string<typename T::value_type>& result) const
+inline void result::result_impl::get_ref_impl(short column, T& result) const
 {
     bound_column& col = bound_columns_[column];
     const SQLULEN column_size = col.sqlsize_;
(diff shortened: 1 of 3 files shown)


nanodbc/nanodbc.cpp
```diff
@@ -3161,6 +3165,37 @@ inline void result::result_impl::get_ref_impl<std::vector<std::uint8_t>>(
...
+void result::result_impl::get_ref_from_string_column(short column, T& result) const
+{
+    bound_column& col = bound_columns_[column];
+    if (col.ctype_ != SQL_C_CHAR && col.ctype_ != SQL_C_WCHAR)
+        throw type_incompatible_error();
+    string str;
+    get_ref_impl(col.column_, str);
+    std::string temp;
+    convert(str, temp);
+    std::istringstream strstr(temp);
```
mloskot: I'd suggest to avoid the streams and use the C++11 `stox` functions with a little bit of indirection, e.g.:

```
namespace detail {
template <typename T>
auto from_string(T const& s, double) { return std::stod(s); }
template <typename T>
auto from_string(T const& s, int) { return std::stoi(s); }
template <typename T>
auto from_string(T const& s, long) { return std::stol(s); }
}

template <typename R, typename T>
auto from_string(T const& s) -> R
{
    return detail::from_string(s, R{});
}
```

then the line in `get_ref_from_string_column` above would read:

```
result = from_string<T>(str);
```

Quick test w/ VS2017 suggests the `stoX` are 10-20 faster

ChrisBFX: Yes this is a good idea, it is not only faster, it would also throw an out_of_range exception if the value is to big (sstream simply returns the biggest possible value). The only problem is that there is no function for short and with a simple cast from stoi you would sometimes throw an out_of_range error and sometimes not but return a wrong value.
In c++17 there is `std::from_chars` which also works for short (though the error handling is strange).
Another solution would be boost::lexical_cast.

mloskot: > there is no function for short and with a simple cast from stoi you would sometimes
> throw an out_of_range error and sometimes not but return a wrong value.

Good point. `from_string(T const& s, short)` would have to get a bit of logic checking `std::stoi` return value fits the range of `short`, I think.

> In c++17 there is std::from_chars 

Yes, but AFAIK, nanodbc 2.x line (also current master) sticks to C++14

> Another solution would be boost::lexical_cast.

Boost should be considered optional, and `short` could be handled with that but only if `NANODBC_ENABLE_BOOST=ON`

nanodbc/nanodbc.cpp
```diff
@@ -3161,6 +3165,37 @@ inline void result::result_impl::get_ref_impl<std::vector<std::uint8_t>>(
...
 
+template <>
+void result::result_impl::get_ref_from_string_column(short column, string::value_type& result) const
+{
+    bound_column& col = bound_columns_[column];
+    const char* s = col.pdata_ + rowset_position_ * col.clen_;
+    switch (col.ctype_)
+    {
+    case SQL_C_CHAR:
+        result = (string::value_type) * (const char*)(s);
```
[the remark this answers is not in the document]
lexicalunit: Yah, we try to avoid C-style casts (even tho C++ casts are indeed long and ugly) because they're more obvious and generally safer (can't accidentally cast away const-ness, for example).

nanodbc/nanodbc.cpp
lexicalunit: `boost::lexical_cast` is kinda neat, I was following the boost mailing lists as that was going through it's development and review. But nanodbc aims to be as compact as possible without reliance on other libraries, no matter how generally accepted and supported they are. If we can use what's standard then we should try to at least.

ChrisBFX: Ok, i've changed the stringstream conversion to the from_string style (it uses stoll/stoull for integer types, and checks if it is inside the limits of the requested type). Can somebody run the tests, travis/appveyor seems not to be working on this repository?

mloskot: The project is in middle of transition to GitHub organization.

I think we will have to wait with the merge, or perhaps we don't? What we do @lexicalunit ?

mloskot:
I'm not sure why, but GitHub did not notify me about "approved these changes" update.

This PR will have to be merged manually into the new https://github.com/nanodbc/nanodbc

mloskot:
Merged https://github.com/nanodbc/nanodbc/commit/2bd4e2e9e8cfa52742d9e641e6b1bca7a3100b94

